PR #1, opened 2015-06-05
Added Travis file to allow CI building on Github.

Danack:
This allows Pull Requests to be built automatically, so that you don't even have to consider merging PRs that break the build.

Apologies if this PR is to the wrong branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: c
+
+
+compiler:
+  - clang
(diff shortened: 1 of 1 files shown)


dlemstra:
We added your file to our SVN repository and added a Markdown file so the status is visible on the start page. It currently breaks so thank you for adding this so we can fix this :+1:

PR #2, opened 2015-06-05
Remove files built by build process.

Danack:
This files get regenerated by ./configure and so I don't believe they should be under source control.


dlemstra:
The files have been removed from our SVN repository.

Danack:
Hi - it appears the file is still in git, but just as a zero-length file. https://github.com/ImageMagick/ImageMagick/blob/master/PerlMagick/Makefile.PL

urban-warrior:
It is required for PerlMagick to build under Windows with Strawberry Perl.  Dirk, may be able to generate it with the Windows configure program-- until then Makefile.PL will remain as an empty file.

dlemstra:
The Windows configure program has been modified and the file has been removed again.

PR #8, opened 2015-06-11
Remove generated files from IM7

Danack:
It appears these files are also auto-generated.


dlemstra:
The files have been removed, thanks!

PR #9, opened 2015-06-11
Add function getHdriEnabled

Danack:
Add function to allow determining if ImageMagick was compiled with HDRI enabled at run time.

This may seem slightly ridiculous as it is obviously available at compile time - the problem it solves is that people seem to install Imagick with an version of ImageMagick from different places, and it's not guaranteed that Imagick was compiled against IM with the same HDRI setting. Adding the function to get the HDRI setting allows it to be checked at run-time.


dlemstra:
I think it would be better to add a new method for this in opencl.h/c.

urban-warrior:
GetMagickFeatures() currently returns a string of features which includes "HDRI" if ImageMagick @ run-time is HDRI-enabled.

PR #14, opened 2015-06-27
First pass at fully escaping all string output in json.c

edalquist:
This pull request works though I'm not sure I'd say I'm proud of the form. Like I said in the original bug C is not my forte. If you have suggestions for improvements please let me know.

I took the very defensive approach and simply json escaped any string printed out by json.c I'd guess some subset of those strings are never going to be invalid but it is probably better to be consistent and safe.

I tested a few images with metadata that is invalid in JSON via:
`convert IMAGEFILE -moments -features 1 json:- | jsonlint`

https://github.com/zaach/jsonlint

### DIFF
--- a/coders/json.c
+++ b/coders/json.c
@@ -388,7 +388,11 @@ static ssize_t PrintChannelFeatures(FILE *file,const ChannelType channel,
   ssize_t
     n;
 
-  n=FormatLocaleFile(file,FeaturesFormat,name,
+  char
+    *escaped_json;
+
+  escaped_json = EscapeJsonString(name);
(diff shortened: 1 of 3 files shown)


dlemstra:
Thanks for the help. I just committed a modified version of your code to our SVN repository. Please let me know if I missed anything or if you have suggestions for improving the JSON output.

edalquist:
Thanks! I just ran this on a new photo set and found that the json escape logic needs to handle control characters too. I'll get a follow up patch some time today.

dlemstra:
When can we expect your new pull request?

PR #20, opened 2015-07-23
Rewrite MagickCore-config.1 man page with propper man macro's.

kwm81:
This started by me noticing "holes" in the manual pages. Now FreeBSD uses mandoc to "format" manual pages, which comes from OpenBSD. While mandoc is stricter it didn't give that many warnings on the old manpage, it just showed holes. mandoc also strives to be compatible with other manual page formatters like groff. 

Rewrite the manpage using "proper" man macro's. I also changed a few things like grouping arguments that did the same thing together. I would like some feedback about the new setup so I can take that into account when I update the other manual pages.

Please add a e-mail at the end or just remove the <e-mail here> part before commit.

### DIFF
--- a/MagickCore/MagickCore-config.1
+++ b/MagickCore/MagickCore-config.1
@@ -1,69 +1,71 @@
-.ad l
-.nh
-.TH MagickCore-Config 1 "2 May 2002" "ImageMagick"
-.SH NAME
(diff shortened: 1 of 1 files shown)


urban-warrior:
Your changes worked fine with nroff so we commited MagickConfig.1.  Feel free to update the other manual pages.  Thanks.

PR #21, opened 2015-07-31
typo in doc

ubante:
Found it earlier today while installing.


dlemstra:
Thank you for reporting this, we have submitted your patch to our SVN repository.

PR #34, opened 2015-09-24
Prevent null pointer access.

Danack:
The change below should prevent any null pointer access. It also needs to be applied to IM7 branch I presume.

Are there any unit tests to cover this functionality? I couldn't find them, so I couldn't run them.

### DIFF
--- a/magick/constitute.c
+++ b/magick/constitute.c
@@ -1273,7 +1273,7 @@ MagickExport MagickBooleanType WriteImages(const ImageInfo *image_info,
     status;
 
   register Image
-    *p;
+    *p, *next_p;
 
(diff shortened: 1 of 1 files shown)


urban-warrior:
ImageMagick has thousands of unit-tests, activate them with 'make check'.

Danack:
Ok, can you point me to the tests that covers the 'scene' functionality and I'll add an extra bit that covers this case?

urban-warrior:
Most of the unit tests are in tests/validate.c.  You can always add a unit test there.  There are other one-off unit tests in the tests folder driven by the TAP scripts.

PR #39, opened 2015-09-25
Improved documentation.

UniQP:


dlemstra:
I cherry picked your fix for the number of columns. Thanks for fixing this for us. But I don't think it is necessary to document the return value if can only be true or false.

UniQP:
Of course it is important to document the return value. The return value can indicate whether an error occurred or whether the execution of the function was successful. Furthermore, the return value is not a boolean but a MagickBooleanType. Thus, it is a good idea to mention the possible return values MagickTrue and MagickFalse (that I need to check the return value).

PR #42, opened 2015-09-27
Corrected type for ExportLongPixel to use long.

Danack:
The change below seems necessary for exportImagePixels to work correctly on Centos 6.4. Without it, the value set in the pixels array are garbage.
 
Oddly the current code seems to work correctly on Ubuntu, which isn't immediately obvious why it works.

### DIFF
--- a/magick/pixel.c
+++ b/magick/pixel.c
@@ -1268,7 +1268,7 @@ static void ExportLongPixel(Image *image,const RectangleInfo *roi,
   register const PixelPacket
     *restrict p;
 
-  register unsigned int
+  register unsigned long int
     *q;
 
   register ssize_t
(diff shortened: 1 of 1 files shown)


urban-warrior:
The *Long* methods are a bit of a misnomer, they import / export 4 byte unsigned ints.  That suggests the current "unsigned int *q" is correct.  Can you provide a use case where it fails under CentOS x64?  We have CentOS x64 and cannot reproduce any failures.

Danack:
Ah, that got me. Can you confirm that my new understanding of the types to use is correct?

CharPixel - unsigned 1 byte integer
DoublePixel - 8 byte floating point
FloatPixel - 4 byte floating point
IntegerPixel - Removed in IM7 unsigned 4 byte integer
LongPixel - unsigned 4 byte integer
QuantumPixel - Quantum, use the defined type from the IM headers.
ShortPixel - unsigned 2 byte integer

urban-warrior:
That's correct.  We'll add support for LongLongPixel for 64 bit unsigned integers.

Danack:
Cool. Thank-you.

PR #56, opened 2015-11-18
Fix build when fchmod is missing

endrift:
fchmod doesn't exist on all platforms, and it's already detected by the configure script, so just don't use it if it's not found.


dlemstra:
Thanks for the patch. We have applied it to IM6 and IM7.

PR #69, opened 2015-12-22
Add complex text support

ibtisamN:
This pull request adds support for languages that require [complex text layout](https://en.wikipedia.org/wiki/Complex_text_layout).

We are using [libraqm](https://github.com/HOST-Oman/libraqm), a small source code-only library that wraps FriBidi (for bidirectional text support) and HarfBuzz (for text shaping), and does proper BiDi and script itemization.

The CTL support is enabled by default but can be disabled at compiling time, and we provide a fallback function that uses your original code without CTL support.

Since we are now doing proper bidirectional text, the existing `-direction` becomes meaningless as it reverses the order of all characters unconditionally, so we changed it to set the base  paragraph direction instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+sudo: required # needed for trusty beta
+dist: trusty   # needed for HarfBuzz
+
 language: c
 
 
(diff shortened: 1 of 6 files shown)


dlemstra:
We have applied a refactored version of your patch to both IM6 and IM7. Thank you for your contribution.

Fahad-Alsaidi:
Hi @dlemstra  I didn't understand how I can enable CTL now ? is there any flag for it?

khaledhosny:
The changes to `configure.ac` as well as the `raqm.{c,h}` files from this PR seem to be missing on master.

PR #70, opened 2015-12-28
Fix kerning in fallback text layout

ibtisamN:
Freetype kerning changes the advance of the previous glyph
instead of the x offset. Also draw_info->kerning was already
applied in renderFreeType, so no need to apply it in the
fallback function.


dlemstra:
We have applied your patches to both IM6 and IM7. Thanks for your contribution.

PR #75, opened 2016-01-07
Resource leak in AcquireVirtualMemory

anorland:
Pair AcquireMagickResource with calls to RelinquishMagickResource when resource limits are hit. Returning early on success makes the error handling easier.

Remove temporary files when extending or mapping a file fails.


urban-warrior:
We applied your patch to the GIT master.  Thank you for your contribution.

PR #87, opened 2016-01-19
Write correct pitch and channel bitmasks for uncompressed DDS images.

boardwalk:
I noticed that images I converted with `convert in -define dds:compressed=none out.dds` didn't have sensible values for pitch and channel bitmasks. This fixes that.


dlemstra:
Thank you for your contribution we have applied your patch to IM7 and IM6.

PR #115, opened 2016-02-01
Fix cluster value in fallback code

ShamsaHamed:
The value of the cluster should be the byte index of the
character not the actual character.


mikayla-grace:
We applied your patch to the ImageMagick-6 branch.  Thanks.

dlemstra:
Can you confirm this issue has been resolved so we can close your pull request?

ShamsaHamed:
It's Ok. Thanks.

PR #137, opened 2016-02-23
Fix multiple DDS output issues.

Hrnchamd:
Fixes errors in encoding logic for DXT5 alpha, which caused alpha blockiness and incorrect alphas if a block hit a certain branch of FixRange.

Fixes DDS header to follow the specification. Writes linear size or stride depending on compression. Should prevent output DDS files from crashing GIMP.

Addresses problems in https://www.imagemagick.org/discourse-server/viewtopic.php?f=3&t=27169&sid=39ea05ff30d1fe5a2ae55ce4d547207a

The original code seems to be converted from libsquish, but suffered some logic errors in conversion. It lacks unit tests comparing the output to libsquish output, which would have caught the alpha problem.

### DIFF
--- a/coders/dds.c
+++ b/coders/dds.c
@@ -728,7 +728,7 @@ if (min > max) \
 if (max - min < steps) \
   max = MagickMin(min + steps, 255); \
 if (max - min < steps) \
-  min = MagickMax(min - steps, 0)
+  min = MagickMax(0, max - steps)
 
 #define Dot(left, right) (left.x*right.x) + (left.y*right.y) + (left.z*right.z)
 
@@ -2689,9 +2689,14 @@ static void WriteDDSInfo(Image *image, const size_t pixelFormat,
     flags;
 
   flags=(unsigned int) (DDSD_CAPS | DDSD_WIDTH | DDSD_HEIGHT |
-    DDSD_PIXELFORMAT | DDSD_LINEARSIZE);
+    DDSD_PIXELFORMAT);
   caps=(unsigned int) DDSCAPS_TEXTURE;
   format=(unsigned int) pixelFormat;
+
+  if (format == DDPF_FOURCC)
+      flags=flags | DDSD_LINEARSIZE;
+  else
+      flags=flags | DDSD_PITCH;
 
   if (mipmaps > 0)
     {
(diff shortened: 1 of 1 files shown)


Hrnchamd:
9c26fe1 seems to have got an incorrect uncompressed pitch calculation. Preparing an update.

dlemstra:
Thanks for your patches. I will try to put them in the master and our IM6 branch tommorow.

dlemstra:
I finally had some time for this patch but now I have some questions about your patch @Hrnchamd. This is about the calculation of dwPitchOrLinearSize. I could find the following in the documentation:

|Flags|Description|
|---|---|
|DDSD_PITCH|Required when pitch is provided for an uncompressed texture.|
|DDSD_LINEARSIZE|Required when pitch is provided for a compressed texture.|

> The pitch or number of bytes per scan line in an uncompressed texture; the total number of bytes in the top level texture for a compressed texture. For information about how to compute the pitch, see the DDS File Layout section of the Programming Guide for DDS.

For uncompressed images and setting the correct flags your patch looks correct. For `DDPF_FOURCC` we should provide the total number of bytes in the top level texture but when you take a look at the page https://msdn.microsoft.com/en-us/library/windows/desktop/bb943991(v=vs.85).aspx it is only the size of one scanline. And according to the first table it expects a pitch. So he documentation is a bit vague on this. Can you clarify/provide a source why you multiply it by the number of rows?

Hrnchamd:
Your linked page is referring solely to the pitch calculation, rather than the header specification. The actual DDS_HEADER spec at https://msdn.microsoft.com/en-us/library/windows/desktop/bb943982%28v=vs.85%29.aspx specifies the correct meaning for dwPitchOrLinearSize. With an incorrect value, GIMP will fail loading the image.

It seems like the DDS summary page was updated for DX10 without referencing the rest of the specification properly, ignoring that dwPitchOrLinearSize semantics depend on the pixel format.

dlemstra:
Your patches have finally been applied. I apologize for taking this long. Would you be willing to also create a patch for our ImageMagick-6 branch @Hrnchamd?

p.s. If you do please include the minor changes that I made.

Hrnchamd:
Ah great. I'll start looking at ImageMagick-6 now.

PR #199, opened 2016-05-14
Fix typo s/width/height/ in resource.c (ImageMagick-6 branch)

mirakui:
Hi there,
I found that `width` is wrongly used instead of `height` in `magick/resource.c`.
Could you check this?


dlemstra:
Thanks for finding this bug and sending us a pull request. Your changes have been merged.

PR #219, opened 2016-06-11
fix cache time limit, fix #217

### LINKED ISSUE #217: Lot of Operation canceled' @ fatal/cache.c/GetImagePixelCache/1885
Version 6.9.4-1 was ok
Observed since 6.9.4-3

See https://github.com/mkoppanen/imagick/issues/158

remicollet:


remicollet:
Perhaps the "if ((cache_timelimit ..." can even be in an else, as on first pass, when cache_epoch is not yet set, doesn't need to be check.

remicollet:
Second try seems simpler, and better for perf.

dlemstra:
We have manually merged your changes, thank you!

PR #223, opened 2016-06-19
Fix abort when writing to rgf format

dlech:
The rgf format (LEGO MINDSTORMS EV3 images) caused a software abort because
exception == NULL. When WriteRGFImage is called from WriteImage, it is only
passed two parameters, not three. So, removed the extra parameter and use
image->exception instead as in other coders.

### DIFF
--- a/coders/rgf.c
+++ b/coders/rgf.c
@@ -68,7 +68,7 @@
   Forward declarations.
 */
 static MagickBooleanType
-  WriteRGFImage(const ImageInfo *,Image *,ExceptionInfo *);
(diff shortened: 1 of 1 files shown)


dlech:
This patch is tested working on Ubuntu 16.04

dlemstra:
Good catch :+1:  Your commit has been cherry picked (https://github.com/ImageMagick/ImageMagick/commit/a0108a892f9ea3c2bb1e7a49b7d71376c2ecbff7) and this fix will be available in the next release (6.9.4-10).

PR #231, opened 2016-07-16
Fixes for mingw-w64

jeroenooms:
These are fixes needed to build on `mingw-w64` platform. The CRT does not expose `_create_locale` and `__free_locale` and also the `#include <intsafe.h>` header is not available and not needed.


dlemstra:
We have cherry picked your changes, thanks for your contribution :+1: We also applied your changes to our ImageMagick 6 branch.

PR #237, opened 2016-07-23
Fix misaligned bits when writing .rfg

dlech:
When writing LEGO MINDSTORMS EV3 .rfg file, if the image is not a
multiple of 8 pixels wide, the pixels in the last byte are not aligned
correctly. The pixels need to be shifted the number of bits that are
not used in the last byte.

### DIFF
--- a/coders/rgf.c
+++ b/coders/rgf.c
@@ -387,7 +387,10 @@ static MagickBooleanType WriteRGFImage(const ImageInfo *image_info,Image *image)
       p++;
     }
     if (bit != 0)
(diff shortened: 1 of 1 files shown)


dlech:
FWIW, I have compiled, tested and done a binary diff with a known good file to confirm that this fix works.

dlemstra:
Thanks for helping us out again :+1: . Your contribution has been cherry picked in both the master and our ImageMagick-6 branch. Thanks for providing us with a patch for both versions.

PR #260, opened 2016-08-30
Fix text wrapping bug

arthurdarcet:
The bug was described here: http://www.imagemagick.org/discourse-server/viewtopic.php?f=3&t=30376

tl;dr:
```convert -size 400x -pointsize 50 caption:'This does not wrap' out.png```

If the last letter of the last word is the only thing overflowing, but there is a wrap point somewhere available, then it should be used to wrap.

If I'm not mistaken, the strcmp check was only here to avoid an infinite loop when trying to render a single word in a too small width. If so, the check i used instead does the job.

### DIFF
--- a/MagickCore/annotate.c
+++ b/MagickCore/annotate.c
@@ -634,7 +634,7 @@ MagickExport ssize_t FormatMagickCaption(Image *image,DrawInfo *draw_info,
     if (status == MagickFalse)
       break;
     width=(size_t) floor(metrics->width+draw_info->stroke_width+0.5);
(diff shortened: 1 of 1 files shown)


dlemstra:
Thanks a lot for the contribution :+1: 

I removed the unused variable 'text' afterwards. That variable was added in patch 74966db to resolve the issue reported here: https://www.imagemagick.org/discourse-server/viewtopic.php?f=3&t=19502. And I am happy to report that with your fix this still works.

arthurdarcet:
Thanks, I completely missed the `text` var  was unused after this patch

PR #311, opened 2016-11-30
fix url missing hyphen

muralisr:
the download links on the binary releases page points to urls that are 404s because they miss the hyphen after libs.


dlemstra:
Thanks for helping us out 👍. The source of the webpages is our Website repo. This was not yet mirrored on github and gitlib but I added the mirrors just now. We have already commited your changed a couple days ago: https://github.com/ImageMagick/Website/commit/3ec33a0ec23831d18080c7bd4885ee1fc357fb6b. If you find something else please create a pull request in the Website repo.

PR #314, opened 2016-12-03
Fix bug with missing return

erw7:
I found an access exception to occur because there is no return in windows ImageMagick 6.
Could you check this?


dlemstra:
Thanks for finding this bug 👍. Your changes have been merged and this bug will be fixed in the next version of ImageMagick 6.

PR #315, opened 2016-12-03
Fix #313

### LINKED ISSUE #313: Magick++ is not compliant to C++1z
Magick++ contains hundreds of syntax errors since [std::unary_function<>](http://en.cppreference.com/w/cpp/utility/functional/unary_function) got removed from the C++ standard. In particular, Visual Studio 2017 RC with the /std:c++latest switch enabled refuses to compile it.  
This is causing our library for artificial intelligence, [Hippocrates](https://github.com/SirRade/Hippocrates), great trouble, as we planned to use Magick++ for it's [image recognition module](https://github.com/IDPA16/CNN-Code).  

Fortunately, the fix is rather simple: Because std::unary_function<> was guaranteed to be an empty base, you can either
- Just remove the unnecessary inheritance completly
- replace it by an empty class called `Magick::unary_function` or, for no replacement, even `Magick::std::unary_function`

This fix should take at most 15 minutes. Should I create a pull request?

SirRade:
#313

### DIFF
--- a/Magick++/lib/Magick++/Drawable.h
+++ b/Magick++/lib/Magick++/Drawable.h
@@ -94,8 +94,7 @@ namespace Magick
   //
   // Base class for all drawable objects
   //
-  class MagickPPExport DrawableBase:
-    public std::unary_function<MagickCore::DrawingWand,void>
+  class MagickPPExport DrawableBase
   {
   public:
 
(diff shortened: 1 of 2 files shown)


dlemstra:
Thanks for the patch. Your commit has been cherry picked into the master branch. Can you also create a pull request for the IM6 branch?

Issue #249, opened 2016-08-12
register' storage class specifier is deprecated


jeroenooms:
```
/usr/local/include/ImageMagick-7/MagickCore/pixel-accessor.h:628:3: warning: 'register' storage class specifier is deprecated [-Wdeprecated-register]
  register ssize_t
```

mikayla-grace:
Thanks for the problem report.  We can reproduce it and will have a patch to fix it in GIT master branch @ https://github.com/ImageMagick/ImageMagick later today.  The patch will be available in the beta releases of ImageMagick @ http://www.imagemagick.org/download/beta/ by sometime tomorrow.

dlemstra:
Can you confirm that this is resolved @jeroenooms ?

PR #358, opened 2017-01-18
Fix compile warning bug for deprecated register

akantsc:
There is stall a compiler warning for Magick++ as discussed in the previous issue:

https://github.com/ImageMagick/ImageMagick/issues/249
Changes for previous issue: https://github.com/ImageMagick/ImageMagick/commit/1e4662d6938e2780173db729b9fc32fd30718180

This pull request adds the register define in a needed core header


dlemstra:
Your changes have been manually merged.

PR #382, opened 2017-02-17
fix fd leak for webp coder

snowwalf:
Signed-off-by: zhangji <zhangji@qiniu.com>

It seams that fd leaks when parse image with format webp.

cmd such as:
convert a.webp -quality 75.0  b.webp

https://github.com/ImageMagick/ImageMagick/blob/master/coders/webp.c +250
The function ReadWEBPImage() will use OpenBlob() to open one fd for source image.
However, no CloseBlob() func was found until the end of ReadWEBPImage()

It leads to fd leak for server program which use imageMagick SDK.


dlemstra:
Merged and added ChangeLog entry. Thank you very much for finding this bug 🙌.

PR #477, opened 2017-05-06
fix DICOM read with fractional slope

jcupitt:
The DICOM reader was multiplying pixel values by rescale_slope, which it
represented as ssize_t. Many DICOM files use double values for
rescale_slope, so for rescale_slope < 1, the reader was multiplying by
zero and producing blank images.

This commit removes all internal handling of rescale_slope and
rescale_intercept, leaving it to the caller to apply these values. It
should be applied to the 7.x branch as well.

This change makes sense within the DICOM standard, since rescale_slope
and rescale_intercept say how to get from the uint16 (or whatever the
file holds) to the value in the units that the file represents, perhaps kBQ/ml.
These are almost always fractional quantities and int-valued pixels
cannot represent them.

Here's a sample file that shows the problem:

http://www.rollthepotato.net/~john/MRIm5.dcm

You're welcome to use this image in a test suite. Before this commit:

$ identify -verbose MRIm5.dcm
  Image: MRIm5.dcm
  Format: DCM (Digital Imaging and Communications in Medicine image)
  Gray:
    min: 32769 (0.500023)
    max: 32769 (0.500023)

After:

$ identify -verbose MRIm5.dcm
  Image: MRIm5.dcm
  Format: DCM (Digital Imaging and Communications in Medicine image)
  Gray:
    min: 0 (0)
    max: 65535 (1)

Forum thread:

https://www.imagemagick.org/discourse-server/viewtopic.php?f=3&t=31887&p=145852

### DIFF
--- a/coders/dcm.c
+++ b/coders/dcm.c
@@ -2877,12 +2877,10 @@ static MagickBooleanType ReadDCMPixels(Image *image,DCMInfo *info,
                   }
                 i++;
               }
-          index=(int) ((pixel_value*info->rescale_slope)+
-            info->rescale_intercept);
           if (info->window_width == 0)
             {
               if (info->signed_data == 1)
-                index-=32767;
+                index=pixel_value-32767;
             }
           else
             {
@@ -2894,14 +2892,14 @@ static MagickBooleanType ReadDCMPixels(Image *image,DCMInfo *info,
                 (info->window_width-1.0)/2.0-0.5);
               window_max=(ssize_t) floor((double) info->window_center+
                 (info->window_width-1.0)/2.0+0.5);
-              if ((ssize_t) index <= window_min)
+              if ((ssize_t) pixel_value <= window_min)
                 index=0;
               else
-                if ((ssize_t) index > window_max)
+                if ((ssize_t) pixel_value > window_max)
                   index=(int) info->max_value;
                 else
-                  index=(int) (info->max_value*(((index-info->window_center-
-                    0.5)/(info->window_width-1))+0.5));
+                  index=(int) (info->max_value*(((pixel_value-
+		    info->window_center-0.5)/(info->window_width-1))+0.5));
             }
           index&=info->mask;
           index=(int) ConstrainColormapIndex(image,(size_t) index);
--- a/coders/dcm.c
+++ b/coders/dcm.c
@@ -2697,8 +2697,6 @@ typedef struct _DCMInfo
     window_width;
 
   ssize_t
-    rescale_intercept,
-    rescale_slope,
     window_center;
 } DCMInfo;
 
@@ -3083,8 +3081,6 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
   info.depth=8;
   info.max_value=255UL;
   info.mask=0xffff;
-  info.rescale_intercept=0;
-  info.rescale_slope=1;
   info.samples_per_pixel=1;
   info.scale=(Quantum *) NULL;
   info.signed_data=(~0UL);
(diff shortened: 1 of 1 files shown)


dlemstra:
Thanks for your patch but I did some research why this patch was introduced this was done because I had an image with a rescale_intercept that was not zero. Can you adjust your patch to keep in the rescale_intercept but leave out the rescale_slope?

jcupitt:
I'm not sure, that seems very non-intuitive to me. As a user of a library, I would expect it to apply slope and intercept, or not to apply slope and intercept. Applying one but not the other feels wrong.

Also, couldn't intercept change the range of the data? -1000 is a common value in CT for images in Hounsfield units, and that would cut off all values <1000, unless the output images was changed to signed short. And changing it to signed short would cut off all values >33767. 

https://en.wikipedia.org/wiki/Hounsfield_scale

Again, the only way to represent the full range of the data safely is to output float32, or perhaps int32. Or to not apply intercept.

dlemstra:
How about we only apply the intercept when the slope has a value that is 1? This was the case with the image where I added the patch for. The output was too bright and applying the intercept resolved that issue.

jcupitt:
That feels like strange behaviour as well. It would need documenting carefully or people would be very puzzled why intercept was sometimes being applied and sometimes not. All application code that uses IM would need changing to add an extra `if()` clause to sometimes apply the intercept and sometimes not. 

Is there some other way your too-bright image could be handled? You could maybe use `-fx` and apply the slope and intercept that way?

In my work I use the IM DICOM loader to get the uint16 image into Python, then I pull out intercept and scale and transform it into a float image for analysis. I process CT, MRI and PET data all in the same way, and it's mostly worked well. Could that kind of data path work for your case?

dlemstra:
You are right that only applying the intercept when the slope is 1 is strange behavior. I am now leaning more towards another solution.

We keep the rescale intercept and slope but make it a double instead. And we will add a new option called 'dcm:ignore-rescale' that you can use to disable the rescale being applied. Most people that use ImageMagick don't do any post processing with metadata on the images so I think it make sense to have rescale enabled as the default.

jcupitt:
I agree, that sounds like a good solution!

I wonder about having it default to off though, and calling the option `dcm:enable-rescale`. Rescale will very often cause data loss and will break programs written for the old behaviour (everything at my University). 

Your call, of course.

dlemstra:
You are right that I should not forget the backwards compatibility issue. Disabling it by default is probably the right way to go. I do think there should be two settings for the option though so I think it might be better to just call it `dcm:rescale` instead. One of those options would be `on`. and the other one would be only applying the `intercept` when the `rescale` value is 1. I am not sure how I would want to call this option, do you have an idea for that?

jcupitt:
Hi again, I came across this somewhat related issue:

http://www.imagemagick.org/discourse-server/viewtopic.php?t=29825

Your problem with too-bright images seems rather like this. `window_center` / `window_width` are supposed to be there for transforming for display, `rescale_slope` and `rescale_intercept` are for getting from the raw uint16 in the file into the proper units for the scan (kBq/ml or time or whatever). 

Reading the DICOM spec (I keep a copy by my bed in case I have difficult sleeping), it seems they should be applied after rescale:

https://www.dabsoft.ch/dicom/3/C.11.2.1.2/

So how about two new options:

1. A boolean-valued option to enable the application of `rescale_slope` and `rescale_intercept`. I don't think you can apply one and not the other, since they have a defined, physical meaning as a pair. Perhaps call it `dcm:rescale`. 

2. If `dcm:rescale` is enabled, then also apply the `window` values stored in the DICOM file, don't apply them otherwise. And add a new option called maybe `dcm:window` that lets you override center and width from the command-line as `center:width` or `center@width` or whatever the best IM syntax might be. 

We could make that user very happy too.

dlemstra:
I do think we already apply the window here: https://github.com/ImageMagick/ImageMagick/blob/master/coders/dcm.c#L2887 and that does not solve the issue with the file I have. I don't think I can share the file that I received from the user so you can see what is happening. But I do think we could now live without the second option. We can always add this at a later point when more users request this. Just enabling the option would be enough for that user now. They don't have files with a rescale_slop that is not `1`.

jcupitt:
Yes, I saw that the window is there. By the spec, you're supposed to only apply the window if you first apply the rescale. 

For your user, they could just enable the rescale. I guess the window is zero for them? If we're adding options, I thin we might as well add a window override as well, it's very little work. 

Shall I close this PR and make a new one with a proposed patch?

dlemstra:
The window is indeed zero. Would be great if you could create a new pull request for this. And I think we should also follow the specs and only apply the window when rescale is enabled. I would like the `dcm:window` option in a new pull request after we have merged the first one. I would rather not mix this. And for syntax you should use centerXwidth (`-define dcm:window=100x1000`).

jcupitt:
OK, new PR opened, I'll close this one.

I don't suppose you know the best way to parse centerXwidth? I'm struggling to find a helper that does exactly what we need.

dlemstra:
You could use the method `ParseGeometry` for this. Our codebase has enough usage examples.

Issue #478, opened 2017-05-07
DrawSetTextInterwordSpacing bogus code


Danack:
[This line](https://github.com/ImageMagick/ImageMagick/blob/7fcca7ff26c0579f3010222b9a44063ba0b45182/MagickWand/drawing-wand.c#L6123) in DrawSetTextInterwordSpacing appears bogus. It should either be calling fabs for `CurrentContext->interword_spacing-interword_spacing` or it could be safer to do it as it's done for the [ImageMagick 6 branch](https://github.com/ImageMagick/ImageMagick/blob/01d522e990aa57cbe67d222dd5e8f7196cc6d199/wand/drawing-wand.c#L6129), where it is a direct comparison without epsilon.

Currently, the code won't work if the new interword_spacing is greated than the old interword_spacing.

This was reported downstream at:

https://github.com/mkoppanen/imagick/issues/206
https://bugs.php.net/bug.php?id=72438

Danack:
The same pattern appears to be affecting [DrawSetTextKerning](https://github.com/ImageMagick/ImageMagick/blob/7fcca7ff26c0579f3010222b9a44063ba0b45182/MagickWand/drawing-wand.c#L6041) and [DrawSetTextInterlineSpacing](https://github.com/ImageMagick/ImageMagick/blob/7fcca7ff26c0579f3010222b9a44063ba0b45182/MagickWand/drawing-wand.c#L6082) .

KevinGregull:
I can confirm that changing the Line to the following does work and fix the Problem:

`if ((wand->filter_off != MagickFalse) &&
      (CurrentContext->interline_spacing!=interline_spacing))
    {`

KevinGregull:
Seeing how they chose to use the Delta Value, I think using the absolute difference is a better solution here.
Created a Pull Request for the change.

KevinGregull:
Thank you for spotting the error @Danack. Fix applied, this Issue can be closed.

dlemstra:
Your changes have been merged and will be available in the next release.

PR #480, opened 2017-05-07
Fixed missing absolute value

KevinGregull:
Fixed bogus code that didn't allow for DrawSetTextInterlineSpacing,DrawSetTextKerning & DrawSetInterwordSpacing to be changed, because of a missing absolute value.

#478


dlemstra:
Thanks for your patch. We have also applied these changes to IM6.

PR #483, opened 2017-05-09
make DICOM read work for fractional rescale

jcupitt:
The DICOM reader was multiplying pixel values by `rescale_slope`, which it
represented as `ssize_t`. Many DICOM files use double values for
`rescale_slope`, so for `rescale_slope` < 1, the reader was multiplying by
zero and producing blank images.

This patch changes `rescale_slope` and `rescale_intercept` to be double, so
dcm read now works with fractional values of rescale.
This patch should be applied to the 7.x branch as well.

This patch incorporates a couple of other small changes too:

* Previous versions of ImageMagick did not apply the rescale
  values, so for compatibility, application of them now defaults to
  off. Use `-define dcm:rescale` to enable rescale processing.

* `window_center` and `window_width` should also be double-valued, from
  the spec.

* The `window_center` and `window_width` fields were always applied before,
  but from the DICOM spec, they should be only be applied to the rescaled
  values.

  The patch makes width and center processing only happen if `dcm:rescale`
  is on.

* Previously, the offset for signed data was only being applied if
  windowing was disabled. It is now always applied.

Here's a sample file that shows the problem:

http://www.rollthepotato.net/~john/MRIm5.dcm

You're welcome to use this image in a test suite. Before this patch:

$ identify -verbose MRIm5.dcm
Image: MRIm5.dcm
Format: DCM (Digital Imaging and Communications in Medicine image)
Gray:
min: 32769 (0.500023)
max: 32769 (0.500023)

After:

$ identify -verbose MRIm5.dcm
Image: MRIm5.dcm
Format: DCM (Digital Imaging and Communications in Medicine image)
Gray:
min: 0 (0)
max: 65535 (1)

Forum thread:

https://www.imagemagick.org/discourse-server/viewtopic.php?f=3&t=31887&p=145852

Related PR:

https://github.com/ImageMagick/ImageMagick/pull/477

### DIFF
--- a/coders/dcm.c
+++ b/coders/dcm.c
@@ -3104,7 +3104,7 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
   info.samples_per_pixel=1;
   info.signed_data=(~0UL);
   info.significant_bits=0;
-  info.rescale=0;
+  info.rescale=MagickFalse;
   info.rescale_intercept=0.0;
   info.rescale_slope=1.0;
   info.window_center=0.0;
@@ -4097,8 +4097,7 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
           }
         option=GetImageOption(image_info,"dcm:rescale");
         if (option != (char *) NULL)
-          info.rescale=(int) ParseCommandOption(MagickBooleanOptions,
-            MagickFalse,option);
+          info.rescale=IsStringTrue(option);
         status=ReadDCMPixels(image,&info,stream_info,MagickTrue,exception);
         if ((status != MagickFalse) && (stream_info->segment_count > 1))
           {
--- a/coders/dcm.c
+++ b/coders/dcm.c
@@ -2693,13 +2693,16 @@ typedef struct _DCMInfo
     max_value,
     samples_per_pixel,
     signed_data,
-    significant_bits,
-    window_width;
+    significant_bits;
 
-  ssize_t
+  MagickBooleanType
+    rescale;
+
+  double
     rescale_intercept,
     rescale_slope,
-    window_center;
+    window_center,
+    window_width;
 } DCMInfo;
 
 typedef struct _DCMStreamInfo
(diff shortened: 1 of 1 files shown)


jcupitt:
If/when this patch is accepted, I'll make another PR which adds `dcm:window` so that the window center and width can be overridden from the command-line.

coders/dcm.c
```diff
@@ -4076,6 +4095,10 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
             if (LocaleCompare(option,"reset") == 0)
               info.window_width=0;
           }
+        option=GetImageOption(image_info,"dcm:rescale");
+        if (option != (char *) NULL)
+          info.rescale=(int) ParseCommandOption(MagickBooleanOptions,
```
dlemstra: You could use `IsStringTrue` here.

coders/dcm.c
```diff
@@ -3080,20 +3094,21 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
...
-  info.rescale_slope=1;
+  info.max_value=255UL;
   info.samples_per_pixel=1;
-  info.scale=(Quantum *) NULL;
   info.signed_data=(~0UL);
   info.significant_bits=0;
-  info.window_center=0;
-  info.window_width=0;
-  info.polarity=MagickFalse;
+  info.rescale=0;
```
dlemstra: This should be MagickFalse instead.

coders/dcm.c
```diff
@@ -2877,32 +2880,43 @@ static MagickBooleanType ReadDCMPixels(Image *image,DCMInfo *info,
...
-              else
-                if ((ssize_t) index > window_max)
-                  index=(int) info->max_value;
-                else
-                  index=(int) (info->max_value*(((index-info->window_center-
-                    0.5)/(info->window_width-1))+0.5));
+              index=pixel_value;
             }
+          if (info->signed_data == 1)
+            index-=32767;
```
dlemstra: The old code did not do this when the `windows_size` was not 0. Is this change intentional?

dlemstra:
We would also welcome a patch for our IM6 branch 👍  (*hint*)

coders/dcm.c
jcupitt: Yes, it seemed correct to me. If the output is signed, you must offset the origin, whether or not you interpret the window. I have not test files I could check with, though.

coders/dcm.c
jcupitt: Oh nice. I copy-pasted that code from the webp loader, perhaps that should change too.

jcupitt:
This patch is for the IM-6 branch, I agree a patch to master would be nice too. Should I make another PR once this is done?

jcupitt:
OK, pushed the changes you proposed. Thanks for the review!

dlemstra:
Feel free to wait with the IM7 patch until we have applied the `dcm:window` changes in IM6.

jcupitt:
\o/ thanks! I'll make another PR for the window stuff.

jcupitt:
`dcm:window` PR is here

https://github.com/ImageMagick/ImageMagick/pull/484

PR #484, opened 2017-05-10
add dcm:window option

jcupitt:
This patch adds a new `dcm:window` option which lets you override the
window center and width stored in the DICOM file with a new centerXwidth
pair. You can also use centerX or Xwidth. The window settings are ony
applied if the `dcm:rescale` option is enabled.

This is useful for transforming images from scan units to display units for
visualisation.

Example:

convert -define dcm:rescale -define dcm:window=2500x3000 MRIm5.dcm x.tif

This patch also reverts an earlier change which moved the transform for
signed data into the main path. Window handling assumes a 0 - `max_value`
output range and takes no account of signedness of the data, so
the subtract for signed data is not appropriate. Perhaps it should be,
but we'd need to track `min_value` as well, and that's beyound the scope
of this PR.

Related PR:

https://github.com/ImageMagick/ImageMagick/pull/483

DICOM spec:

https://www.dabsoft.ch/dicom/3/C.11.2.1.2/

### DIFF
--- a/coders/dcm.c
+++ b/coders/dcm.c
@@ -2880,6 +2880,8 @@ static MagickBooleanType ReadDCMPixels(Image *image,DCMInfo *info,
                   }
                 i++;
               }
+          if (info->signed_data == 1)
+            pixel_value-=32767;
           if (info->rescale)
             {
               double
@@ -2914,8 +2916,6 @@ static MagickBooleanType ReadDCMPixels(Image *image,DCMInfo *info,
           else
             {
               index=pixel_value;
-              if (info->signed_data == 1)
-                index-=32767;
             }
           index&=info->mask;
           index=(int) ConstrainColormapIndex(image,(size_t) index);
@@ -4093,15 +4093,12 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
             if (LocaleCompare(option,"reset") == 0)
               info.window_width=0;
           }
-        option=GetImageOption(image_info,"dcm:rescale");
-        if (option != (char *) NULL)
-          info.rescale=(int) ParseCommandOption(MagickBooleanOptions,
-	    MagickFalse,option);
         option=GetImageOption(image_info,"dcm:window");
         if (option != (char *) NULL)
 	  {
             GeometryInfo
               geometry_info;
+
             MagickStatusType
               flags;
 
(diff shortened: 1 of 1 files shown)


coders/dcm.c
```diff
@@ -4097,7 +4095,24 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
...
         if (option != (char *) NULL)
-          info.rescale=IsStringTrue(option);
+          info.rescale=(int) ParseCommandOption(MagickBooleanOptions,
+	    MagickFalse,option);
+        option=GetImageOption(image_info,"dcm:window");
+        if (option != (char *) NULL)
+	  {
+            GeometryInfo
+              geometry_info;
+            MagickStatusType
```
dlemstra: This needs a newline between the variables :)

coders/dcm.c
```diff
@@ -4097,7 +4095,24 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
           }
         option=GetImageOption(image_info,"dcm:rescale");
         if (option != (char *) NULL)
-          info.rescale=IsStringTrue(option);
```
dlemstra: You should really use IsStringTrue here.

coders/dcm.c
```diff
@@ -4097,7 +4095,24 @@ static Image *ReadDCMImage(const ImageInfo *image_info,ExceptionInfo *exception)
...
+          info.rescale=(int) ParseCommandOption(MagickBooleanOptions,
+	    MagickFalse,option);
+        option=GetImageOption(image_info,"dcm:window");
+        if (option != (char *) NULL)
```
dlemstra: I wonder if we should make it so that if this option is not null we set info.rescale to MagickTrue. This option will only do something when info.rescale is true. And the you would only need to do this: `convert -define dcm:window=2500x3000 MRIm5.dcm x.tif`

coders/dcm.c
jcupitt: OK, if you're sure.

dlemstra: I'm sure 👍

dlemstra:
Your changes have been merged (and I added the IsStringTrue check). Could you also send the pull request for IM7?

jcupitt:
Ooop, sorry, I did make that change, I think I forgot to push it. 

I'll try an IM7 patch that does the same thing.

jcupitt:
I meant to say, I didn't write a note in the ChangeLog, nor any docs or tests. Do they need doing too?

dlemstra:
Forgot to mention that you should add something to the change log. And our website is in another repo. We have a page where we explain all the defines. Would love it if you could add something there for the two new options that you created.

jcupitt:
I added a note to the ChangeLog in this branch, hopefully you can just merge that too.

I added some docs to the website.

IM7 next!

jcupitt:
IM7 PR: https://github.com/ImageMagick/ImageMagick/pull/485

dlemstra:
I don't see the push for that ChangeLog change?

jcupitt:
I pushed it, but it looks like I have to open a new PR for github to see it. Hang on.

PR #516, opened 2017-06-21
Link zlib (-lz) in delegates tests.

egzumer:
When building delegates from source (with --enable-delegate-build), TIFF fails tests as there is no -lz parameter in compiler call.


mikayla-grace:
Thanks for the problem report.  We can reproduce it and will have a patch to fix it in GIT master branch @ https://github.com/ImageMagick/ImageMagick later today.  The patch will be available in the beta releases of ImageMagick @ http://www.imagemagick.org/download/beta/ by sometime tomorrow.

dlemstra:
Your patch has been manually merged. Thanks 👍

PR #607, opened 2017-07-24
Fix underlinking in DNG module

antonio-rojas:
Use correct variable for libraw libs. Should fix https://www.imagemagick.org/discourse-server/viewtopic.php?f=3&t=32367


dlemstra:
"Manually merged"

PR #622, opened 2017-07-26
fix gif loop iterations

clyfish:
Value 0 in "NETSCAPE2.0" means loop forever.
Missing "NETSCAPE2.0" block means 0 repetitions = loop once.
Value 1 in "NETSCAPE2.0" means 1 repetition = 2 loops.
Value 2 in "NETSCAPE2.0" means 2 repetitions = 3 loops.

https://bugs.chromium.org/p/chromium/issues/detail?id=592735
https://bugzilla.mozilla.org/show_bug.cgi?id=1255497
https://stackoverflow.com/questions/27066511/how-to-make-an-animated-gif-loop-twice-in-imagemagick


dlemstra:
We have manually merged your commit. Thanks for your contribution!

PR #635, opened 2017-07-30
./coders/bmp.c:1723 says it is bmp3:alpha

MarkSetchell:


dlemstra:
Thanks for the patch but you need to send a pull request in this repo: https://github.com/ImageMagick/Website/blob/master/include/command-line-options.php.

MarkSetchell:
Hi Dirk,

Sorry I created the pull request the wrong way. Do you have a minute to explain why it was wrong please? The website files seem to be duplicated both repos which I don't understand.

I am pretty unfamiliar with "git" so go easy on me please!

Many thanks
Mark

dlemstra:
Of course I will go easy on you. This repository contains a copy of the www files. The original files are located in the `Website` repository. If you would change it here it would be overwritten the next time we update the files.

PR #681, opened 2017-08-18
bug fix about word-spacing

jiangzhenhong:
should use '||' instead of '&&' to check when set word-spacing


dlemstra:
Thanks for the fix @jiangzhenhong 👍

PR #749, opened 2017-09-12
Add .gitattributes file

rcdailey:
I'm going to assume everyone knows what a `.gitattributes` file is, so I won't go over that here. If not please ask.

The main benefit to having this (especially the `* text=auto` part) is repository-centric handling of EOL characters and conversions, which completely supersede the `core.eol` and `core.autocrlf` settings, which are a nightmare to setup properly depending on platform.

I followed [github's recommended process for normalizing line endings](https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings) in the repository, and I didn't see any conversions needed.

Please let me know which files must explicitly maintain LF line endings on Windows platform, because that is the only thing that needs to be explicitly set up in the `.gitattributes` file. For example, *.sh scripts usually won't work on Windows with CRLF line endings (when executing them through mingw).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# Let Git determine what is a text file and handle line ending normalization
+* text=auto
+
+# Files on Windows that should retain LF line endings
+*.sh        eol=lf
+Makefile*   eol=lf


dlemstra:
Could you clarify a bit more what this solves? We haven't had any newline issue.

rcdailey:
It's not solving any particular problem at the moment. It does prevent inconsistent line endings based on differing user configuration in Git across platforms. For example, without this, your repository can potentially get a mixture of CRLF and LF endings in source files, unless you make sure all users have proper configuration.

Having this ensures that your *repository* determines the EOL conversion process and expected line endings in the repository itself. This keeps things within your control.

If you're not familiar with line ending mumbo jumbo in Git, I recommend you do some reading on it. I don't mean for that to sound like I'm saying RTFM, quite literally it's a complex and nasty problem. Git attributes file solves it for you though.

dlemstra:
We have had no issues setting up the line handling on our machines but that might be because we have a really small team. And I have also never heard that it is a nightmare to set up. But that doesn't mean we won't consider applying your changes if that would help you. Our repository also contain a couple of binary images. Should we not mark them as binary in the `.gitattributes` file or does `auto` know how to handle jpg/png/ico/gif images?

rcdailey:
Does anyone on your team develop on Windows? Or are you all on Linux/Mac? If the latter, you likely wouldn't ever run into issues. Again, things start getting messy when CRLF is thrown into the mix, which implies Windows is a development platform by someone on your team.

I'm doing this to help you guys, not really myself. Since your repository is open to public contribution, you don't want to have to worry about making sure line endings look correct in pull requests prior to merging. The `.gitattributes` will fix that for you.

Concerning binary files: Yes, auto should know not to muck with image files. But you handle these on a case-by-case: If you notice on checkout of a PNG file that it gets corrupted, you simply add an explicit line for it in the `.gitattributes`:

    *.png binary

It's more maintainable to handle the exceptional cases rather than everything. But it doesn't hurt, I think it comes down to personal preference. My preference is to trust git as much as possible, because that means less work I have to do.

randolf:
@rcdailey Should the following files also be included for LF preservation?

- AppRun
- configure
- configure.ac
- Makefile.am
- Makefile.in

dlemstra:
And we prob also need to mark the images as binary files.

rcdailey:
@randolf: It doesn't hurt to add them, but I do find a lot of linux utilities are able to handle CRLF endings on Windows. Some of the MINGW tools can for example. I have found that bash/sh do not handle CRLF in shell scripts, which is why I almost always add `eol=lf` to those in the `.gitattributes`. You can play around with makefiles and configure if you want to be conservative.

@dlemstra: In my experience, Git is very good at detecting binary files. EOL processing does not happen to binary files when `* text=auto` is specified. This tells git to detect binary files automatically WRT eol handling. Personally, I only add binary files to `.gitattributes` if I notice it's not being detected as binary by Git. So to answer your question, I would say images do not need to be explicitly marked right now unless you are noticing a problem.

PR #751, opened 2017-09-13
bugfix: (thread-safty) the id would be changed before it is returned. 

liclicli:
the code in wand/wand.c is not thread-safe. The id would be changed before it returned. That would cause memory leaking and performance drop down after a long time running. 

### Prerequisites


dlemstra:
Thanks for this really nice find! 🥇 👍 🙌. Could you make the following change?
```
size_t nID = 0;
```

To:
```
size_t
  wand_id=0;
```

PR #752, opened 2017-09-13
fix memory leak in ReadYCBCRImage as SetImageExtent failure

jgj212:
Here is critical code:
```
  //...

  quantum_info=AcquireQuantumInfo(image_info,canvas_image);
  if (quantum_info == (QuantumInfo *) NULL)
    ThrowReaderException(ResourceLimitError,"MemoryAllocationFailed");

 //...

    status=SetImageExtent(image,image->columns,image->rows,exception);
    if (status == MagickFalse)
      return(DestroyImageList(image));    //should free the memory about quantum_info

//...


Credit: ADLab of Venustech

### DIFF
--- a/coders/ycbcr.c
+++ b/coders/ycbcr.c
@@ -212,6 +212,7 @@ static Image *ReadYCBCRImage(const ImageInfo *image_info,
     if (status == MagickFalse)
     {
       quantum_info=DestroyQuantumInfo(quantum_info);
+      canvas_image=DestroyImage(canvas_image);
       return(DestroyImageList(image));
     }
     SetImageColorspace(image,YCbCrColorspace,exception);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dlemstra** [comment]
Thanks for this pull request 👍 . I wonder if you should also add `canvas_image=DestroyImage(canvas_image);` to destroy that image.

**Comment by jgj212** [comment]
@dlemstra yes, you are right. I will fix it soon

**Comment by dlemstra** [comment]
Thanks for the quick response. Could you check if we also have this issue in the IM6 branch and open a pull request there?

**Comment by dlemstra** [comment]
And could you write the if like this:
```c
if ()
  {
// instead of
if ()
{
```

nohmask:
This was assigned CVE-2017-15032.

PR #754, opened 2017-09-14
fix multiple memory leak in ReadYUVImage

jgj212:
We found that there is multiple memory leak vulnerability in ReadYUVImage  because no memory free on exception condition. And we create patch here.

Credit: ADLab of Venustech

### DIFF
--- a/coders/ycbcr.c
+++ b/coders/ycbcr.c
@@ -212,6 +212,7 @@ static Image *ReadYCBCRImage(const ImageInfo *image_info,
     if (status == MagickFalse)
     {
       quantum_info=DestroyQuantumInfo(quantum_info);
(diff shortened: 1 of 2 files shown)


dlemstra:
Thanks for helping us 👍. It seems that your changes now have conflicts. Can you rebase your changes to resolve this? And could you make sure that you are using the correct indentation? Might be incorrect because you are using tabs instead of spaces but I cannot see that now. We use 2 spaces for indentation.

dlemstra:
And it looks like you also included your `ycbcr.c` fixes in this pull requests again. But with different indentation.

jgj212:
@dlemstra  i will resolve conflicts, thank you

jgj212:
I create a pull request, so closed this

PR #756, opened 2017-09-14
fix multiple memory leak in ReadYUVImage

jgj212:
We found that there is multiple memory leak vulnerability in ReadYUVImage because no memory free on exception condition. And we create patch here.

Credit: ADLab of Venustech


dlemstra:
Thanks for this pull request could you also make one fro the IM6 branch?

jgj212:
@dlemstra i will test im6 later

nohmask:
This was assigned CVE-2017-15033.

PR #787, opened 2017-09-28
fix multiple null pointer

jgj212:
fix multiple null pointer


coders/art.c
```diff
@@ -325,6 +325,8 @@ static MagickBooleanType WriteARTImage(const ImageInfo *image_info,Image *image,
   (void) WriteBlobLSBShort(image,0);
   (void) WriteBlobLSBShort(image,(unsigned short) image->rows);
   quantum_info=AcquireQuantumInfo(image_info,image);
+  if(quantum_info==NULL)
```
dlemstra: Could you change this to: `if (quantum_info == (QuantumInfo *) NULL)`? And also include the type for all the other checks?

jgj212: Ok, i will change all of them

dlemstra:

dlemstra:
Have you checked if we also have these issues in IM6? Could you create a new pull request to fix the mistakes in IM6?

PR #788, opened 2017-09-29
fix multiple null pointer 

jgj212:
fix multiple null pointer

### DIFF
--- a/MagickCore/cache.c
+++ b/MagickCore/cache.c
@@ -3836,8 +3836,6 @@ MagickExport MagickBooleanType PersistPixelCache(Image *image,
     Clone persistent pixel cache.
   */
   clone_info=(CacheInfo *) ClonePixelCache(cache_info);
-  if (clone_info == (CacheInfo *) NULL)
-    return(MagickFalse);
   clone_info->type=DiskCache;
   (void) CopyMagickString(clone_info->cache_filename,filename,MagickPathExtent);
   clone_info->file=(-1);
--- a/MagickCore/cache.c
+++ b/MagickCore/cache.c
@@ -3836,6 +3836,8 @@ MagickExport MagickBooleanType PersistPixelCache(Image *image,
     Clone persistent pixel cache.
   */
   clone_info=(CacheInfo *) ClonePixelCache(cache_info);
+  if (clone_info == (CacheInfo *) NULL)
+    return(MagickFalse);
   clone_info->type=DiskCache;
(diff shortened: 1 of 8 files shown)


MagickCore/cache.c
```diff
@@ -3836,6 +3836,8 @@ MagickExport MagickBooleanType PersistPixelCache(Image *image,
     Clone persistent pixel cache.
   */
   clone_info=(CacheInfo *) ClonePixelCache(cache_info);
+  if (clone_info == (CacheInfo *) NULL)
```
dlemstra: We don't need this check. `ClonePixelCache` calls `AcquirePixelCache` that will throw a fatal exception when the pixel cache could not be acquired.

MagickCore/stream.c
```diff
@@ -1290,8 +1290,15 @@ MagickExport Image *StreamImage(const ImageInfo *image_info,
   assert(stream_info->signature == MagickCoreSignature);
   assert(exception != (ExceptionInfo *) NULL);
   read_info=CloneImageInfo(image_info);
+  if (read_info == (ImageInfo *) NULL)
```
dlemstra: `CloneImageInfo` will never return null because the first call is to `AcquireImageInfo` that will throw a fatal exception when it fails.

MagickCore/stream.c
```diff
@@ -1290,8 +1290,15 @@ MagickExport Image *StreamImage(const ImageInfo *image_info,
...
+    return (Image *) NULL;
   stream_info->image_info=image_info;
   stream_info->quantum_info=AcquireQuantumInfo(image_info,(Image *) NULL);
+  if (stream_info->quantum_info == (QuantumInfo *) NULL)
```
dlemstra: `AcquireQuantumInfo` will throw a fatal exception when it fails so we also don't need this check.

dlemstra: My bad!! We need this check it is possible that it returns null!

MagickCore/stream.c
jgj212: You are right.  I reviewed the code, I don't know why i made this  mistake.

MagickCore/cache.c
jgj212: You are right.

MagickCore/stream.c
dlemstra: @jgj212 Can you restore this check. I did not check this correctly. It is possible that `AcquireQuantumInfo` returns null.

Issue #854, opened 2017-11-05
[Package Request] Please provide Linux binary packages in AppImage format


KurtPfeifle:
I noticed that ImageMagick has tagged 41 different releases of the 7.0.x variant in 2017 alone. That is almost weekly a new release.

It would very nice and extremely welcome for users who want to get their hands on the latest and greatest of ImageMagick, if you also could provide binaries in the  [AppImage](https://appimage.org/) format, and if you could configure the build process which you already have automated with [Travis](https://travis-ci.org/ImageMagick/ImageMagick/jobs/297613215) to spit out such a beast.

> ###  TL;DR
>
> An AppImage has the following advantages:
>
> 1. Directly runs on all major Linux distribution.
> 1. Similar functionality as "PortableApps" on Windows.
> 1. Directly runs from any location (even USB thumbdrives).
> 1. You can run different versions of the same App side-by-side.
> 1. No need to "install".
> 1. An AppImage is a single file.
>
> Technically, an AppImage is posing as a normal ELF binary to the outside world. Inside, the most part of it is a compressed [AppDir](http://rox.sourceforge.net/desktop/AppDirs.html) (see [also Wikipedia](https://en.wikipedia.org/wiki/Application_directory)) containing the application alongside the required run-time libraries and resources in the shape of an ISO file system tree. A small piece of extra code inside the AppImage auto-mounts the AppDir on a mount point in *`/tmp/`*.
>
> Then it executes the binary inside the AppImage.

So, an AppImage is a file, which the end user can directly start (well, after making it executable), from any location (even USB thumb drive), across many different distros, old and new.

Actually, AppImages aren't even "packages" in the same sense as other software packages. They don't need an "install" step, as one is used to for other software. They bundle everything that is required into that one single file, which makes the "package management" as easy as managing simple files.

The most basic principle of the AppImage format is: `"1 file == 1 application"`. Contrary to what the first reading of their website may lead one to believe (*"So this is for Desktop / GUI applications only?"*), AppImages work very well for CLI tools too.
<sub>*(I've successfully tested AppImages made from the `ippsample` software, provided by the ISTO Printer Working Group [PWG] for prototyping and testing IPP Everywhere [Internet Printing Protocol] software.)*</sub>.

The `"1 file == 1 application"` makes it very easy for users to make use of the software:

1. Download + save 1 file only.
1. Make it executable.
1. Run it (from whereever you stored it).

[...]

PR #858, opened 2017-11-09
Generate AppImage on Travis CI

probonopd:
### Prerequisites

### Description
This PR, when merged, will convert the existing [Travis CI](https://travis-ci.org/) builds into an [AppImage](http://appimage.org/) and uploads it to the GitHub Releases page.

Closes https://github.com/ImageMagick/ImageMagick/issues/854.

Providing an [AppImage](http://appimage.org/) will have, among others, these advantages:
- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions
- Can be listed in the [AppImageHub](https://appimage.github.io/apps) central directory of available AppImages

[Here is an overview](https://appimage.github.io/apps) of projects that are already distributing upstream-provided, official AppImages.

__Please note:__ For the builds to be uploaded to GitHub Releases, you need to set `GITHUB_TOKEN` as a secure environment variable in Travis CI; also see https://github.com/probonopd/uploadtool.

If you have questions, AppImage developers are on #AppImage on irc.freenode.net.


dlemstra:
This sounds like a good plan but we do have a couple questions. Would it be possible to only do an AppImage build when we add a tag? And you might want to build with `MAGICKCORE_ZERO_CONFIGURATION_SUPPORT` enabled so you can avoid the version number inside the AppRun file. Otherwise we would need to change it from time to time.

probonopd:
> Would it be possible to only do an AppImage build when we add a tag?

Sure, but imho that would take away from the benefit of having continuous builds readily available for testing. Be aware that you will only have one `continuous` release, as on the next build this tag will be removed and re-created. Also, continuous builds are available even for Pull Requests (although only as temporary download URLs in the build log and not shown on the GitHub Releases page).

But if you want to change this behavior, it should not be a problem. Just let me know.

> And you might want to build with `MAGICKCORE_ZERO_CONFIGURATION_SUPPORT` enabled

Thanks for the hint @dlemstra - does this mean I can drop the

```
export MAGICK_HOME="$HERE/usr" # http://www.imagemagick.org/QuickStart.txt
export MAGICK_CONFIGURE_PATH="$HERE/usr/lib/ImageMagick-7.0.7/config-Q16/" # Undocumented?
```

dlemstra:
We do all our work in the master branch so having a new AppImage for each commit would not be something we would want. It is possible that we are committing work in progress and have something published that is not done yet. I don't understand what you mean but removing and re-creating the `continuous` tag.

You shouldn't need to set those paths with `MAGICKCORE_ZERO_CONFIGURATION_SUPPORT` but have not used it myself ever so I you might need to test if it works.

probonopd:
> I don't understand what you mean but removing and re-creating the continuous tag.

Maybe an example is easier to explain: Looking at 
https://github.com/AppImage/AppImageUpdate/releases
you will only see one release, "continuous", which is also marked as "Pre-Release", since that project has not released a real release yet. On each `git push` the content of "coninuous" gets replaced with the build artefacts of the latest Travis CI run, if the build succeeded.

In addition to that, of course releases will show up there once they are made.

> It is possible that we are committing work in progress and have something published that is not done yet.

If you really don't want the continuous AppImages to be uploaded, that's simple to change. I just want to understand your needs. Do you just not want them to show up on the GitHub Releases page but would be fine with a temporary download link buried deep inside the Travis CI build log, or do you really want to avoid having AppImages uploaded for continuous builds entirely?

dlemstra:
I would prefer to not have a continuous AppImage uploaded. I don't mind that someone could download the AppImage from a link inside the Travis CI build log but I would only want an upload to the GibHub releases page to happen when we add a release tag.

dlemstra:
@KurtPfeifle With zero configuration only the build-in configuration will be used. If you would like the user to have their own configuration you should copy the `.xml` files from `config` into the path that you define. You might still want to get rid of the version number though.

probonopd:
>  With zero configuration only the build-in configuration will be used.

If I understand it correctly, that is fine for @KurtPfeifle but he wants to be able to see which options were used at compile time.

KurtPfeifle:
For the benefit of other readers of this thread, to whom it may not be obvious: the latest AppImage builds for ImageMagick7 triggered by @probonopd are here:

* https://github.com/probonopd/ImageMagick/releases

probonopd:
> For the benefit of other readers of this thread, to whom it may not be obvious: the latest AppImage builds for ImageMagick7 triggered by @probonopd are here:
>
>    https://github.com/probonopd/ImageMagick/releases

...not anymore. The download URL for non-tagged builds is now inside the Travis CI build log.

dlemstra:
@KurtPfeifle Those commands now longer work because there are no `.xml` files for zero configuration. I am now rethinking my suggestion and I think we should not use zero configuration but use the `.xml` files so users are able to change their configuration. I think the `export MAGICK_CONFIGURE_PATH=$(readlink -f "$HERE/usr/lib/ImageMagick-*/config-*/") # Undocumented?` looks like a proper fix.

dlemstra:
Do we still need this?:
```
branches:
  except:
    - # Do not build tags that we create when we upload to GitHub Releases
    - /^(?i:continuous)$/
```

dlemstra:
I wonder if we should only publish the release for either the gcc or the clang build. Your thoughts @probonopd?

probonopd:
> I wonder if we should only publish the release for either the gcc or the clang build. 

I don't know, I just assumed you are currently building both for a reason. I don't think it hurts to have both.

dlemstra:
Would you be so kind to remove the last 4 lines that are commented out? They moved from the correct position but I don't think we need them anymore.

dlemstra:
@probonopd We created a couple tags but they are not showing up as downloads on the releases page. Is there anything else we need to do?

probonopd:
Yes, seemingly the `"$TRAVIS_BRANCH" != "TRAVIS_TAG"` is also true for for tagged builds.

https://docs.travis-ci.com/user/environment-variables/ says:

> TRAVIS_BRANCH:
> for builds triggered by a tag, this is the same as the name of the tag (TRAVIS_TAG).

So I'm not sure why this is not working as intended.

dlemstra:
Have you tried to check if `$TRAVIS_TAG` is set?

PR #943, opened 2018-01-14
Fixes for the encoder fuzzer so it runs properly

alex:
Previously the macro invocation was incorrect so it was looking for a decoder named `"FUZZ_IMAGEMAGICK_ENCODR"`. Once that was resolved, adding a `char` to a `char *` was doing pointer arithmetic, not what we wanted.

### DIFF
--- a/Magick++/fuzz/encoder_fuzzer.cc
+++ b/Magick++/fuzz/encoder_fuzzer.cc
@@ -5,9 +5,10 @@
 
 #include "utils.cc"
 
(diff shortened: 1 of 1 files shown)


alex:
Travis failures appear unrelated.

dlemstra:
@alex Is this a fix for: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=5310?

alex:
@dlemstra no, this was a fix for these not working at all -- previously we'd be invoking `magick()` with `"FUZZ_IMAGEMAGICK_ENCODR"`

PR #946, opened 2018-01-15
build the fuzzers in travis

reaperhulk:
This will build the fuzzers in Travis so we have some CI testing on them before oss-fuzz attempts to build/use them.


reaperhulk:
Okay, this now builds against all the fuzzer files (we only build one encoder fuzzer since we're interested in whether it will build and link and have no plans to execute it here)

reaperhulk:
I believe libFuzzer only works on clang due to some clang specific flag. We could separate the fuzzer targets out to a different job inside the build though if you'd like. We'll have to explicitly declare the build matrix but that's only a few lines of yaml.

dlemstra:
Would it be possible to build this without `# Checkout and build libFuzzer`? Maybe we could add a file called `travis.cc` to the folder with the following content:
```C
//#if BUILD_TRAVIS
int main()
{
  return LLVMFuzzerTestOneInput(0,0);
}
//#endif // BUILD_TRAVIS
```

And include this at the bottom of all the `_fuzzer.cc` files?

dlemstra:
And when can we move `build.sh` to our repo? Can't we just add the `build.sh` script to our `Magick++\fuzz` folder?

reaperhulk:
That would almost certainly work. I'll test in a bit.

I'll move build.sh after this PR 😄

reaperhulk:
No luck with the main trick so I just wrapped the whole libFuzzer stuff in a clang block. Now it will only execute on the clang side at least.

dlemstra:
@reaperhulk I had it building on Windows. Did you add `#include travis.cc` as the last line of the file and pass the build argument?

reaperhulk:
Ugh, I wasn't paying attention and assumed you had grabbed that as a travis default var. My mistake. `TRAVIS=true` is actually defined in all travis builds so we can use that. Let me give that a shot.

reaperhulk:
My previous comment was nonsense. I plead jet lag. Anyway, I've pushed a version that works via the include.

PR #959, opened 2018-01-24
add attribute to get in perl doc

fheyer:
Attribute "magick" can be used to get image format tag.

This attribute is important e.g. to determine file type of unknown file.


dlemstra:
@fheyer Thanks for helping out. Is this a get only 'property' or can you also set it?

dlemstra:
Thanks @fheyer 👍

PR #980, opened 2018-02-09
Fix originGeometry JSON value generation

### LINKED ISSUE #979: Bad JSON near `originGeometry` key
### Description
The JSON output option for the `convert` command produces invalid JSON in certain circumstances (according to RFC 7159) and causes errors when trying to parse it.

### Steps to Reproduce
Using an image that has a page origin that is not (0, 0) the `originGeometry` line will have the format `originGeometry: +123+123,` which is invalid. The data should be wrapped in quotation marks. [See line 1452 of json.c][1]

```
$ convert inputFile.jpg json:
```

### System Configuration
I've seen this in Linux and OS X on recent versions of v7.

[1]:https://github.com/ImageMagick/ImageMagick/blob/master/coders/json.c#L1452

tpett:
This should fix #979

### DIFF
--- a/coders/json.c
+++ b/coders/json.c
@@ -1449,7 +1449,7 @@ static MagickBooleanType EncodeImageAttributes(Image *image,FILE *file,
       (double) image->page.width,(double) image->page.height,
       (double) image->page.x,(double) image->page.y);
   if ((image->page.x != 0) || (image->page.y != 0))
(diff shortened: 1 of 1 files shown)


dlemstra:
Thanks for the patch. Would you be so kind to also send us a pull request for the ImageMagick-6 branch?

PR #984, opened 2018-02-13
Revert "Fix previous image references in OptimizeImageLayers"

urban-warrior:
Reverts ImageMagick/ImageMagick#983.  We will need to investigate further.  Setting 0 for the columns or width in CloneImage() returns an image of the same size as the original image.  But notice, the original parameters is set to the page width and height which is not always the same as the image width and height.


akihikodaki:
The original parameters were a bit wrong. The cloned images are used by `CompareImagesBounds`, which just does not care about the page. It should use `columns` and `rows` rather than `page.width` and `page.height`
However, I did not say the change fixes the "wrong" value because those values were still expected to work. `page` should be large enough to contain the image as far as I know, and therefore it was just waste of memory, not affecting the result.

PR #986, opened 2018-02-15
Fixed bit operations for flags.

yoyap:
Most atom boxes head 4 bytes are: { version (1byte) | flags (3bytes) }
-  flags = 0xffffff;
+  flags &= 0xffffff;


dlemstra:
I think it would be wiser to just remove those statements? The flags are not used?

yoyap:
Surely.
However, the program seems to be under development, and I'll withdraw this PR as there may be some intention to put this variable.

PR #987, opened 2018-02-17
Do not refer to page in OptimizeLayerFrames

akihikodaki:
### Prerequisites

### Description
page dimension was used to create images to compare different disposal methods. However, functions used for the comparisons, namely `CompareImagesBounds` and `IsBoundsCleared` are based on the dimensions of images themselves.

Creating image with page dimension is inefficient when converting images with larger pages and therefore should be avoided.

This is what I should have done before #983. See https://github.com/ImageMagick/ImageMagick/pull/984#issuecomment-365310862.

### DIFF
--- a/MagickCore/layer.c
+++ b/MagickCore/layer.c
@@ -987,8 +987,7 @@ static Image *OptimizeLayerFrames(const Image *image,
   /*
     Initialise Previous Image as fully transparent
   */
-  prev_image=CloneImage(curr,curr->page.width,curr->page.height,
-    MagickTrue,exception);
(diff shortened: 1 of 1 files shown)


urban-warrior:
Optimizing images are based on the virtual page layout reflected in the image->page member.  See the GIF image specification for details.  There is likely a bug in optimizing images in some cases but replacing the virtual page with the actual image dimensions will break the optimization algorithm.

akihikodaki:
That should not be problematic because the input is assumes coalesced. We may have assertion if not sure.

urban-warrior:
We'll need a few days to test your assumptions before we can merge the pull.  Stand by...

PR #1033, opened 2018-03-22
add libtiff and libde265 to the oss-fuzz build scripts

reaperhulk:
This will enable TIFF and HEIF fuzzing. We can add more libs (like freetype, etc) as we go.

refs https://github.com/google/oss-fuzz/pull/1263

### DIFF
--- a/Magick++/fuzz/build.sh
+++ b/Magick++/fuzz/build.sh
@@ -1,6 +1,23 @@
 #!/bin/bash -eu
 
-./configure --prefix="$WORK" --disable-shared --disable-docs
+# Build libtiff
+pushd "$SRC/libtiff"
+./autogen.sh
+./configure --enable-static --disable-shared --prefix="$WORK"
+make -j$(nproc)
+make install
+popd
+
+# Build libde265
+pushd "$SRC/libde265"
+./autogen.sh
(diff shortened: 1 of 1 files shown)


dlemstra:
I wonder if we should clone those libraries in our own scripts. Maybe a `clone.sh` script? And I would prefer to clone the copies that we have for the Windows build. We won't link against the latest development code then.

reaperhulk:
The OSS-Fuzz project prefers the data to be cloned during container generation and generally wants development versions so that if issues are found in dependencies they can be fixed and those fixes will quickly propagate to the fuzzer.

dlemstra:
I think it would be nice if we could avoid the PR in the oss-fuzz repo when we add another delegate. Can we not clone the dependencies in a script that is inside the ImageMagick repository?

And how will issues that happen in the other repository be managed? Do we need to contact the owners of the other repository to get this fixed? What would the procedure become?

reaperhulk:
OSS-Fuzz PRs are typically merged extremely quickly, but we can ask if there's a specific reason for the dockerfile logic or if it's just convention.

We will be responsible for contacting upstreams if bugs are found in the dependencies.

PR #1057, opened 2018-03-26
add openjp2, webp, zlib, bzip2, freetype, and libjpeg-turbo to the delegate list for oss-fuzz

reaperhulk:
This PR adds openjpeg, webp, zlib, bzip2, freetype, and libjpeg-turbo to the delegates for the oss-fuzz builds (and likely needs #1056 to avoid additional UBSAN issues)

The reason bzip2 isn't a clone is because it appears bzip2 has no public source control so the source tarball is our only option 😮

### DIFF
--- a/Magick++/fuzz/build.sh
+++ b/Magick++/fuzz/build.sh
@@ -1,5 +1,21 @@
 #!/bin/bash -eu
 
+# build bzip2
+pushd $SRC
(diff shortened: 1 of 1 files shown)


reaperhulk:
This depends on https://github.com/google/oss-fuzz/pull/1270 because I added libjpeg-turbo and freetype2 to the delegate list as well.

PR #1075, opened 2018-03-31
add configure support for disabling utilities with --disable-utilities

reaperhulk:
This PR adds a configure flag (`--with-utilities=yes/no`) that allows us to avoid building the utilities when they are not needed.

I have not rebuilt the Makefile.in and configure scripts from these sources because I have a different version of autotools/automake than what was used to generate the ones in version control and I don't want to clutter the diff.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -782,6 +782,12 @@ AC_ARG_ENABLE(legacy-support,
               [with_legacy_support='no'])
 AM_CONDITIONAL(LEGACY_SUPPORT, test "$with_legacy_support" != 'no')
 
+# Enable building command line utilities (default yes)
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -4006,6 +4011,7 @@ matches your expectations.
   ------------------------------------------------------------------------------
   Shared libraries  --enable-shared=$enable_shared		$libtool_build_shared_libs
   Static libraries  --enable-static=$enable_static		$libtool_build_static_libs
+  Build utilities                               $with_utilities
```
dlemstra: Do we need the --disable-utilities flag here?

PR #1094, opened 2018-04-16
add lcms to the delegate list for oss-fuzz

reaperhulk:
Depends on https://github.com/google/oss-fuzz/pull/1318

This is the last delegate I plan to add for now, but if you'd like more let me know which ones you're interested in and I'll take a look 😄


dlemstra:
Thanks for making this possible 👍

PR #1099, opened 2018-04-17
added HEIF writer and replaced HEIF reader with a more complete HEIF implementation

farindk:
Added code for writing HEIF files based on libheif.
https://github.com/strukturag/libheif
There is a Debian package for libheif, but you will need the latest git version for the encoder (will be released soon).

I also replaced the HEIF reader with the implementation from libheif as the old implementation was limited to only reading HEIF files generated by iPhones. The new implementation is a full implementation of the HEIF standard (still image part) including all image compositing operations.

Alpha channel support is in libheif, but not yet implemented here, since I could not figure out yet how to implement this in ImageMagick.


farindk:
Oh, that's what the 'ping' flag is for. I was wondering about that already
but could not find out the purpose of that flag.

I am a complete beginner to the ImageMagick code, so I do not know yet:
- whether I am doing things right for "lists of images"
- how to handle alpha channels
- what I should do when the HEIF contains several images (libheif supports
that)
- the iPhone also saves "depth images" into the HEIF as an auxiliary image.
How could I pass them to ImageMagick?
- could I pass more options to the HEIF encoder, like whether it should
generate thumbnails, or whether to switch to lossless mode?

Any pointers to information would be appreciated. Especially alpha channels
would be nice, I think.

Dirk


2018-04-17 17:24 GMT+02:00 Dirk Lemstra <notifications@github.com>:

farindk:
Here is support for the 'ping' flag. Hope I did it right.

dbtsai:
Thanks for this PR. This is a great feature supporting HEIF write.

coders/heic.c
```diff
@@ -581,15 +589,31 @@ static MagickBooleanType WriteHEICImage(const ImageInfo *image_info,Image *image
                                                 heif_compression_HEVC,
                                                 &heif_encoder);
     if (error.code) {
+      (void) ThrowMagickException(exception,GetMagickModule(),CoderWarning,
+                                  error.message,"`%s'",image->filename);
       goto error_cleanup;
     }
 
+
+    if (image->quality != UndefinedCompressionQuality) {
```
dlemstra: You should use image_info->quality instead.

farindk:
The libheif package with encoder (v1.1.0) is now in debian testing: https://tracker.debian.org/pkg/libheif
It is also in Ubuntu 18.04.

dlemstra:
@farindk Thats great news!!! 🙌 👍 Could you resolve the conflicts in this branch so I can merge this Friday? We will run autoconf ourselves so maybe you could exclude some of your changes.

dlemstra:
@farindk It looks like we need to use x265 to get encoding support? That won't be an option for us since that is licensed under the GPL v2 license. We could still use the libheif library if that means we would get better support for reading files. Can you remove the writer for now?

farindk:
I was not aware that the ImageMagick license does not allow any GPL v2 libraries, as the ImageMagick license page says that it is compatible to GPL v3. It is because of v2/v3 or because of linking against GPL in general?

How about this solution: I could change the encoding code in libheif in such a way that it does not link to libx265 at all, but uses the separate "x265" program for encoding, passing the data through a pipe. That way, libheif would be LGPL, if I am correct.
In the long term, my plan is to use libde265 itself for encoding instead of x265, but the encoding API is not ready yet.

dlemstra:
We try to avoid linking against GPL libraries. If we also want to get this building under Windows that would mean we would also need to build the x265 library. But I don't want to invest any time in getting that building on Windows due to the license that is being used. The `MAGICKCORE_HEIC_DELEGATE` define is for both the reader and the writer.

While writing the message I just got another idea. I could disable writing with the `MAGICKCORE_WINDOWS_SUPPORT` define under Windows and still enable support for writing under other platforms. I am still happy that your long term plan is getting rid of the x265 dependency. But aren't you forced to use GPL instead of LGPL for the encoder because you are using x265?

farindk:
Disabling it only for Windows sounds like a good solution to me. Then we can easily enable it later when the dependency is removed.

libheif does not require linking to x265. One can register any number of encoding plugins even during runtime. x265 is only one option that is provided by default. But one can also add encoders for AV1, h264, or JPEG. If libx265 is not detected during configuration, that plugin is switched off. In that sense libheif itself is LGPL, but when used together with x265, the combination is GPL.

dbtsai:
@farindk 

I'm using 
```
# convert --version
Version: ImageMagick 7.0.8-10 Q16 x86_64 2018-08-15 https://www.imagemagick.org
Copyright: © 1999-2018 ImageMagick Studio LLC
License: https://www.imagemagick.org/script/license.php
Features: Cipher DPC HDRI Modules 
Delegates (built-in): bzlib freetype heic jng jpeg ltdl lzma png tiff xml zlib
```
with heic support.

When I converted TIFF files with AdobeRGB profile to HEIC files, the color will be changed. `convert DSC01108.tif DSC01108.heic`

Seems the color profile is not properly copied to the new HEIC files. Also, the input file has Colorspace RGB while the output HEIC one is YCbCr. Is it possible to keep as it? How do I specify the compression rate and using 12bit color depth? 

Thanks.

farindk:
Support for color profiles was just added to libheif (https://github.com/strukturag/libheif/pull/58), but it still needs to be integrated to ImageMagick.

RGB colorspace and HDR is on my to-do list. See also https://github.com/strukturag/libheif/issues/62.

dbtsai:
Thanks! I also ran into lossless mode issue. 

FYI, I found `compare` will give huge error due to comparing image in YCbCr and RGB directly without converting YCbCr to RGB.

PR #1102, opened 2018-04-21
Add support for dicts for fuzzers, and added a dict for mvg

alex:


reaperhulk:
This LGTM. Once this is merged we can add a pile of external dictionaries on the oss-fuzz side as well to improve coverage across the current targets.

PR #1103, opened 2018-04-21
reorder build to make tiff happy

reaperhulk:
With the current build order the IM configure script refuses to link against libtiff for reasons unknown. By building libjpeg and libtiff early we work around whatever is going on there. Investigating further would probably be a good idea, but ¯\_(ツ)_/¯

### DIFF
--- a/Magick++/fuzz/build.sh
+++ b/Magick++/fuzz/build.sh
@@ -1,5 +1,19 @@
 #!/bin/bash -eu
 
+# Build libjpeg-turbo
+pushd "$SRC/libjpeg-turbo"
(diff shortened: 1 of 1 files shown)


reaperhulk:
The commit you pushed appears to allow it to build shared libs? That's problematic because the linker will preferentially use the shared objects, but OSS-Fuzz does not support that. We need the whole build to be static so that each fuzzer is a self-contained binary.

dlemstra:
I also statically link libtiff and ImageMagick in my Magick.NET build and it works without that option. I think it doesn't work now because it also need libz when -ltiff is used. I hope that removing this option will fix our check in configure.

reaperhulk:
Closing in favor of #1106

PR #1142, opened 2018-05-26
Fix pdf:page-direction flag

anolivetree:
ViewerPreferences<</PageDirection/R2L>> is inserted into a PDF file when option value is NOT "right-to-left".


dlemstra:
Thanks for the fix 👍

Issue #1143, opened 2018-05-27
More Pixel Art Upscaling Methods


drummyfish:
Hello,

I've been unsuccessfully looking for a tool that implements pixel art scaling methods and so I thought I could write one and integrate it into ImageMagick. AFAIK it currently supports scale2x via [-magnify](https://www.imagemagick.org/script/command-line-options.php#magnify) option, is that right? I'd like to ask:

- Should I? I.e. isn't anyone already working on this, is there an existing tool that I missed etc.?
- What should the CLI options look like? Maybe `-magnify [method]` where `method` could be `eagle2x`, `hq3x` etc.?

The closest thing to what I wanted that I found was [ImageResizer](https://forums.rpgmakerweb.com/index.php?threads/image-resizer-specially-for-upscaling-pixel-art.45856/) but it's written in .NET and doesn't run under mono/wine for me. I'd also prefer a CLI tool. The source of ImageResizer could still be useful to rewrite the algorithms from.

Sorry if I'm posting in a wrong place, I'm new.

urban-warrior:
ImageMagick currently only supports scale2x.  If you want to implement other methods, you will need a new command line option, perhaps -pixel-art method.  If you need pixel art options, utilize defines.  Something like -define pixel-art:threshold=50%.

drummyfish:
Then it would perhaps be best to keep the `-magnify` option and use the `-define` to specify the algorithm? Would make sense.

urban-warrior:
Agreed, something like -define magnify:method=scale2X  -magnify.

PR #1152, opened 2018-05-29
[DO NOT MERGE YET] Add More Pixel Art Upscaling Algorithms

drummyfish:
I am trying to add more pixel art upscaling algorithms as discussed [here](https://github.com/ImageMagick/ImageMagick/issues/1143). This PR is work in progress, I am opening it early to get potential feedback as I am new to ImageMagick and may need some guidance.

Upscaling with pixel art Scale2x algorithm is already supported by IM via `-magnify` option, but only with scale2x algorithm. After this PR I'd like to have an option to specify one of the algorithms such as HQ2x, eagle, scale3x etc. I have also [created my own algorithm](https://github.com/mgba-emu/mgba/pull/737) for MGBA emulator and could add it.

The algorithm should be specified with `-define`, e.g.:

~~`convert -magnify -define "magnify:alg=eagle" in.png out.png`.~~

`convert in.png -define "magnify:method=eagle" -magnify out.png`.

Feel free to write me feedback here. I'll fix the code formatting before merging, I'm not used to this particular style.

EDITS/Notes:

- ~~Comparing pixels based on intensity, as currently implemented, may lead to bugs as two pixels can have the same intensity but different colors, in which case they should probably be considered different.~~ <-- addressed
- Make sure lines aren't longer than 80 chars.
- While some of the algorithms (XBR, ...) usually use pixel comparison with YUV specified error, I am currently comparing pixels for exact values only. The results are basically the same, but may be different for some cases (such as photos). Is this OK or should I stick to the more complicated YUV behavior? If so, the error value might possibly be specified with another -define.
- XBR seems to be a little different from the reference (rewritten from [here](https://github.com/carlosascari/2xBR-Filter/blob/master/xbr.js), reference [here](https://upload.wikimedia.org/wikipedia/commons/c/ce/Pixel-Art_Scaling_Comparison.png)).
- Behavior with indexed-color PNGs that have an alpha channel is buggy (happens also before my changes, IM exits with `Aborted`). `SetImageColorspace` to RGB fixes this but needs to make a copy of the source image.

implemented:

- eagle2x
- HQ2x
- XBR2x
- scale2x (EPX)
- EPBX2x
- eagle3x
- eagle3xb
- scale3x
- fish (my algorithm mentioned above)

### DIFF
--- a/MagickCore/resize.c
+++ b/MagickCore/resize.c
@@ -2044,13 +2044,27 @@ void CopyPixel(const Quantum *src, size_t src_off, Quantum *dst, size_t dst_off,
     dst[channels*dst_off+i] = src[src_off*channels+i];
 }
 
+void MixPixels(const Quantum *src, size_t *src_offs, size_t src_offs_size, Quantum *dst, size_t dst_off, size_t channels)
+{
+  int sum;
+
+  register ssize_t
+    i, j;
+  
+  for (i = 0; i < channels; i++)
+  {
+    sum = 0;
+
+    for (j = 0; j < src_offs_size; j++)
+      sum += src[src_offs[j] * channels + i];
+
+    dst[channels*dst_off+i] = sum / src_offs_size;
+  }
+}
+
 void Mix2Pixels(const Quantum *src, size_t src_off1, size_t src_off2, Quantum *dst, size_t dst_off, size_t channels)
 {
-  register ssize_t
-    i;
-
-  for (i = 0; i < channels; i++)
-    dst[channels*dst_off+i] = (src[src_off1*channels+i] + src[src_off2*channels+i]) / 2;
+  MixPixels(src,(size_t[2]){src_off1,src_off2},2,dst,dst_off,channels);
 }
 
 int PixelsEqual(const Quantum *src1, size_t off1, const Quantum *src2, size_t off2, size_t channels)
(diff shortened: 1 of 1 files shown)


dlemstra:
@drummyfish Looks like a good start 👍  Could you make sure that your lines are not larger than 80 chars?

drummyfish:
> Could you make sure that your lines are not longer than 80 chars?

I definitely will when I get to fixing the formatting part.

drummyfish:
So I have a bunch of methods implemented now. I also tried to fix the formatting, though I probably missed some stuff. Here is what the results currently look like:

image

I'll need some more feedback and code review etc - I'm not really a C programmer so the code will probably need polishing. Also what should I do next? Should I update some documentation somewhere etc.? Can the helper functions stay where they are?

Thank you for all your help.

hummeleBop:
Any news about this upcoming feature ?

GithubSuckBigTime:
Hello, this is OP here. I've left GitHub, which is indicated on my profile, and have long lost track of this feature.

I'd be very glad if someone could finish this though. As far as I remember this was already working, and I've already used it in practice a few times, so there may even be no extra work required than to review and merge, but I don't guarantee this. Feel free to close this PR and create a new one, copy-paste the code or anything, I don't care about any credit or whatever, would just be shame if this was lost I guess.

hummeleBop:
Thanks for your answer.
I can't handle that by myself but it would be amazing.

Too bad, good luck for your next projects :)

Issue #1232, opened 2018-07-31
Orientation info from HEIC decoder does not match actual pixel orientation


rudchenkos:
### Description

It turns out that in HEIC the pixels are never rotated and the metadata only indicates the orientation of the sensor when the photo was taken. After [`ReadHEICImage` populates EXIF profile](https://github.com/ImageMagick/ImageMagick/blob/master/coders/heic.c#L212) and returns a decoded image, the `exif:Orientation` property is [interpreted by ReadImage](https://github.com/ImageMagick/ImageMagick/blob/master/MagickCore/constitute.c#L668) which sets `Image->orientation` accordingly.

Hence, there is a discrepancy between `Image->orientation` and actual pixel orientation in images loaded with the `heic` coder. Particularly, this leads to incorrect behavior of the `-auto-orient` which in this case tries to compensate non-existing pixel rotation.

### Steps to Reproduce

1. Take a HEIC image with `exif:Orientation` not equal to "1" (`TopLeftOrientation`). For instance, a portrait photo with one of the newer Apple iPhone devices. You can download an example photo here: https://www.dropbox.com/s/celi7ejhit9ybzb/test.heic?dl=0
1. Desktop image viewers like `Preview` on macOS normally display the photo as if camera was not rotated.
1. `convert test.heic -auto-orient test.jpg`

#### Expected result

`test.jpg` appears in viewers the same way as original.

#### Actual result

`test.jpg` appears sideways.

### System Configuration
- ImageMagick version: 7.0.8-9 Q16
- Environment: macOS 10.13.6, but the problem is cross-platform

dlemstra:
I get the error below when I try to read your test image. What are the versions of your delegates?

```
I:\heic>identify test.heic
identify: Invalid input: No 'ftyp' box `test.heic' @ error/heic.c/IsHeifSuccess/121.
```

PR #1233, opened 2018-07-31
Align pixel orientation with exif:Orientation in HEIC decoder

rudchenkos:
It turns out that in HEIC the pixels are never rotated and the metadata only
indicates the orientation of the sensor when the photo was taken.

Hence, there was a discrepancy between `Image->orientation` and actual pixel
orientation in images loaded with the `heic` coder. Particularly, this led to
incorrect behavior of the `-auto-orient` which in this case tried to compensate
non-existing pixel rotation.

This is a fix for https://github.com/ImageMagick/ImageMagick/issues/1232

The fix provides two solutions for the problem described above: reset `exif:Orientation` to "1" in HEIC decoder output (default behavior) or rotate pixels to _match_ the original `exif:Orientation`, as some application might rely on the original sensor rotation. The latter is enabled by `heic:preserve-orientation` read define.

### DIFF
--- a/coders/heic.c
+++ b/coders/heic.c
@@ -86,8 +86,6 @@ static MagickBooleanType
   WriteHEICImage(const ImageInfo *,Image *,ExceptionInfo *);
 #endif
 
-static Image *CompensateOrientation(Image *image, ExceptionInfo *exception);
-
 /*
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %                                                                             %
@@ -122,6 +120,55 @@ static MagickBooleanType IsHeifSuccess(struct heif_error *error,Image *image,
     return(MagickTrue);
 
   ThrowBinaryException(CorruptImageError,error->message,image->filename);
+}
+
+/* An inverse of AutoOrientImage */
+static Image *CompensateOrientation(Image *image, ExceptionInfo *exception)
+{
+  const char
+    *value;
+
+  value=GetImageProperty(image,"exif:Orientation",exception);
+  if (value == NULL)
+  {
+    return image;
+  }
+
+  Image
+    *new_image = image;
+
+  switch((OrientationType) StringToLong(value))
+  {
+    case UndefinedOrientation:
+    case TopLeftOrientation:
+    default:
+      return image;
+
+    case TopRightOrientation:
+      new_image = FlipImage(image,exception);
+      break;
+    case BottomRightOrientation:
+      new_image = RotateImage(image,180.0,exception);
+      break;
+    case BottomLeftOrientation:
+      new_image = FlopImage(image,exception);
+      break;
+    case LeftTopOrientation:
+      new_image = TransverseImage(image,exception);
+      break;
+    case RightTopOrientation:
+      new_image = RotateImage(image,270.0,exception);
+      break;
+    case RightBottomOrientation:
+      new_image = TransposeImage(image,exception);
+      break;
+    case LeftBottomOrientation:
+      new_image = RotateImage(image,90.0,exception);
+      break;
+    }
+
+    DestroyImageList(image);
+    return new_image;
 }
 
 static Image *ReadHEICImage(const ImageInfo *image_info,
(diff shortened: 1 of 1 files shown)


rudchenkos:
A question to the maintainers: where in the docs is the best place to describe the new option `heic:preserve-orientation`?

dlemstra:
Is it not enough to call `SetImageProperty(image,"exif:orientation","1",exception);` after the image has been read? The exif profile will be updated before the image is written with the new value.

rudchenkos:
@dlemstra Absolutely. In most common cases it is really enough to reset `exif:Orientation` after the image is read. This is is the default behavior after my patch.

Unfortunately, there are application which rely on the original pixel rotation for historical reasons. This is why I have also implemented the option `heic:preserve-orientation`.

coders/heic.c
```diff
@@ -85,6 +86,8 @@ static MagickBooleanType
   WriteHEICImage(const ImageInfo *,Image *,ExceptionInfo *);
 #endif
 
+static Image *CompensateOrientation(Image *image, ExceptionInfo *exception);
```
dlemstra: Could you move this method above the `ReadHEICImage` method, near `IsHeifSuccess`? We can then remove this line.

coders/heic.c
```diff
@@ -299,8 +305,58 @@ static Image *ReadHEICImage(const ImageInfo *image_info,
...
   heif_context_free(heif_context);
+
+  /*
+    There is a discrepancy between EXIF data and the actual orientation of
+    image pixels. ReadImage processes "exif:Orientation" expecting pixels to be
+    oriented accordingly. However, in HEIF the pixels are NOT rotated.
+   */
+  option=GetImageOption(image_info,"heic:preserve-orientation");
+  if (IsStringTrue(option) == MagickTrue)
+    image = CompensateOrientation(image, exception);
```
dlemstra: This statement leaks memory. The methods `FlipImage`, `RotateImage` return a new image.

coders/heic.c
```diff
@@ -299,8 +305,58 @@ static Image *ReadHEICImage(const ImageInfo *image_info,
...
+  else
+    SetImageProperty(image, "exif:Orientation", "1", exception);
+
   return(GetFirstImageInList(image));
 }
+
+/* An inverse of AutoOrientImage */
+static Image *CompensateOrientation(Image *image, ExceptionInfo *exception)
```
dlemstra: `Image *image` should be `const Image *image`.

coders/heic.c
rudchenkos: Of course, will do 👍

coders/heic.c
rudchenkos: Oh, my. That's right, I will free the old image here

coders/heic.c
rudchenkos: Makes sense, I will change it to a const pointer type

dlemstra: Good point. I missed that, probably better to not change the signature to a const.

rudchenkos:
Hi @dlemstra,
I have applied the suggestions in the fixup commit (to be squashed).

About the docs, it seems the most appropriate place to document `heic:preserve-orientation` is [command-line-options.html](https://github.com/ImageMagick/ImageMagick/blob/master/www/command-line-options.html) right after `hough-lines:accumulator` description. One thing is confusing me though, that line (`www/command-line-options.html:84`) is **very** long which makes me think it is generated by a tool, rather than written manually. But, from the other hand I cannot find where from and how could it have been generated.

Should I just edit that `command-line-options.html`?

dlemstra:
You should edit the command line options here: https://github.com/ImageMagick/Website

coders/heic.c
```diff
@@ -299,6 +341,26 @@ static Image *ReadHEICImage(const ImageInfo *image_info,
...
+    There are two solutions to this problem: either reset the EXIF Orientation tag
+    so it matches the orientation of pixels, or rotate the pixels to match EXIF data.
+   */
+  option=GetImageOption(image_info,"heic:preserve-orientation");
+  if (IsStringTrue(option) == MagickTrue)
+    {
+      Image
+        *original_image = image;
+      image = CompensateOrientation(original_image, exception);
+      DestroyImageList(original_image);
```
dlemstra: You should handle this inside the `CompensateOrientation` method. You are now destroying the `image` when `exif:Orientation` is not set.

rudchenkos: Oh, true. Fill fix promptly.

rudchenkos: Pushed.
I am sorry for the silly mistake

rudchenkos:
Pull request for docs: https://github.com/ImageMagick/Website/pull/19

rudchenkos:
Dank u wel, Dirk!

PR #1252, opened 2018-08-19
Fix compiler warning about unused variable

ralt:
### Prerequisites

### Description

Some values are put in the variable, but the variable isn't ever used afterwards.

```
MagickCore/blob.c: In function 'WriteBlob':
MagickCore/blob.c:5535:6: warning: variable 'q' set but not used [-Wunused-but-set-variable]
     *q;
```

### DIFF
--- a/MagickCore/blob.c
+++ b/MagickCore/blob.c
@@ -5531,9 +5531,6 @@ MagickExport ssize_t WriteBlob(Image *image,const size_t length,
   register const unsigned char
     *p;
 
-  register unsigned char
-    *q;
-
   ssize_t
     count;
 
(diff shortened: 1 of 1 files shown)


ralt:
Nevermind, the compiler doesn't enter the `MAGICKCORE_BZLIB_DELEGATE` define, so it doesn't see the code running in it.

We can't really add an `#if defined()` around the variable declaration either, given that the not-in-defined `case BlobStream` is memcpying into `q`.

Closing this, sorry for the bother.

ralt:
Sorry, my tired mind figured out how to do this right after closing the PR. Reopening with a hopefully good enough commit :-)

ralt:
Hm... looks like github doesn't like it if the branch is force-pushed while the PR is closed. I have opened https://github.com/ImageMagick/ImageMagick/pull/1254 instead.

PR #1253, opened 2018-08-19
Fix compiler warning about unused function.

ralt:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/ImageMagick/ImageMagick/pulls) open
- [x] I have verified that I am following the existing coding patterns and practices as demonstrated in the repository.

### Description

When neither of those compiler macros are available, the compiler is going to spit a warning about the unused static function.

```
coders/miff.c:164:14: warning: 'AcquireCompressionMemory' defined but not used [-Wunused-function]
 static void *AcquireCompressionMemory(void *context,const size_t items,
```

Fix this by making sure the static function is only defined when it's going to be used.

You can easily reproduce the warning with this if you have docker:

```
docker run -ti debian:stretch bash -c 'git clone https://github.com/ImageMagick/ImageMagick && apt-get update && apt-get install autoconf && cd ImageMagick && ./configure && make'
```

PR #1254, opened 2018-08-19
Fix compiler warning about unused variable

ralt:
### Prerequisites

### Description

Some values are put in the variable, but the variable isn't ever used afterwards.

```
MagickCore/blob.c: In function 'WriteBlob':
MagickCore/blob.c:5535:6: warning: variable 'q' set but not used [-Wunused-but-set-variable]
     *q;
```

Put variable declaration and usages in appropriate cpp guards.

### DIFF
--- a/MagickCore/blob.c
+++ b/MagickCore/blob.c
@@ -5531,8 +5531,10 @@ MagickExport ssize_t WriteBlob(Image *image,const size_t length,
   register const unsigned char
     *p;
 
+#if defined(MAGICKCORE_ZLIB_DELEGATE) || defined(MAGICKCORE_BZLIB_DELEGATE)
   register unsigned char
     *q;
+#endif
 
   ssize_t
     count;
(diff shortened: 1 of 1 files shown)


ralt:
Speaking of: #1253 also has `MAGICKCORE_LZMA_DELEGATE`, is it normal that `WriteBlob` isn't caring about it? I'm way too far from understanding what's going on to be able to know if it's normal or not, but that sounds like something worth mentioning.

dlemstra:
This PR looks incorrect to me the `q` variable is not unused. This is used inside the `BlobStream` part that does not depend on the zip or bzip delegate. Feel free to close the PR yourself.

And to answer your question, we support the following streams:

```
typedef enum
{
  UndefinedStream,
  FileStream,
  StandardStream,
  PipeStream,
  ZipStream,
  BZipStream,
  FifoStream,
  BlobStream
} StreamType;
```

Only zip and bzip are supported, lzma is not supported.

urban-warrior:
The q variable is used in the BlobStream case which is not conditionally compiled.  As such, the warning is a false positive or the blob.c was modified recently to correct for this warning.

PR #1271, opened 2018-08-28
travis: test on ppc64le, gcc-8 and a few more library dependencies

grooverdan:
updates m4/ax_gcc_archflag.m4 and the autoreconf changed a bunch of other work.

Migrated existing package install to 'addon: apt: packages'.

Added a make check on ppc64le and added as many dependencies in ubuntu-16.04 that I could find.


dlemstra:
We are open for extra build configurations but our compile only checks if the build succeeds and it doesn't run the unit tests. Could you remove the testing and the extra dependencies to reduce the build time? This is now 17 minutes compared to the 3 minutes for the other configurations.

grooverdan:
ok. Removed as requested. I was just hoping to provide arch testing that isn't always available, despite the increased time. Thanks for the prompt response and considering this merge.

grooverdan:
you're right. it will at least compile test. Fix coming...

dlemstra:
We are planning to move to GitHub actions CI and leave travis. We have a daily build and maybe this could be added there instead?

PR #1278, opened 2018-08-28
use the gcc_archflag for $CXXFLAGS

grooverdan:
Makes the assumption that what works for $CC works for $CXX. (which is why I've set CXX to clang++).


dlemstra:
Closing this as we are planning to move to GitHub Actions CI.

grooverdan:
Are you open to taking the first commit here so that the small C++ component of ImageMagic uses the same architecture optimizations as the C component?

PR #1282, opened 2018-08-30
MagickCore/constitute.c: resolve possible null pointer dereference

chipitsine:
found by cppcheck

[MagickCore/constitute.c:1074] -> [MagickCore/constitute.c:1071]: (warning) Either the condition 'image!=(Image*)NULL' is redundant or there is possible null pointer dereference: image


dlemstra:
Good catch!

PR #1284, opened 2018-08-30
resolve several null pointer dereferences

chipitsine:
those issues were found by cppcheck

### DIFF
--- a/MagickCore/montage.c
+++ b/MagickCore/montage.c
@@ -163,9 +163,9 @@ MagickExport MontageInfo *CloneMontageInfo(const ImageInfo *image_info,
 */
 MagickExport MontageInfo *DestroyMontageInfo(MontageInfo *montage_info)
 {
+  assert(montage_info != (MontageInfo *) NULL);
   if (montage_info->debug != MagickFalse)
     (void) LogMagickEvent(TraceEvent,GetMagickModule(),"...");
-  assert(montage_info != (MontageInfo *) NULL);
   assert(montage_info->signature == MagickCoreSignature);
   if (montage_info->geometry != (char *) NULL)
     montage_info->geometry=(char *)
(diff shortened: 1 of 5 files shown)


MagickCore/nt-base.c
```diff
@@ -1793,20 +1793,20 @@ MagickPrivate DIR *NTOpenDirectory(const char *path)
     {
```
dlemstra: We should remove the null check here instead. `AcquireCriticalMemory` will never be null.

coders/msl.c
```diff
@@ -552,7 +552,8 @@ static void MSLPushImage(MSLInfo *msl_info,Image *image)
   ssize_t
     n;
 
-  (void) LogMagickEvent(TraceEvent,GetMagickModule(),"%s",image->filename);
+  if (image != (Image *) NULL)
```
dlemstra: Good catch!

MagickCore/montage.c
```diff
@@ -412,9 +412,9 @@ MagickExport Image *MontageImageList(const ImageInfo *image_info,
   number_images=GetImageListLength(images);
   master_list=ImageListToArray(images,exception);
   image_list=master_list;
```
dlemstra: Can you move this line down also?

MagickCore/nt-base.c
chipitsine: if I will do it in new commit, can you squash all commits during merge ?

chipitsine:
I pushed with fixes

MagickCore/nt-base.c
dlemstra: I can squash your changes.

dlemstra:
Could you also open a PR for our [ImageMagick6](https://github.com/ImageMagick/ImageMagick6) repository to resolve these issues?

PR #1287, opened 2018-09-01
MagickCore/draw.c: check against null is not needed here

chipitsine:


dlemstra:
Could you please clarify why the null check is not needed?

PR #1288, opened 2018-09-02
remove redundant assignments

chipitsine:


dlemstra:
Could you please split this PR? The redundant assignments part takes a bite more time to make sure we don't break anything.

chipitsine:
all redundant assignments are made within the same function, like that

https://github.com/ImageMagick/ImageMagick/blob/master/Magick%2B%2B/lib/Options.cpp#L801
https://github.com/ImageMagick/ImageMagick/blob/master/Magick%2B%2B/lib/Options.cpp#L804

those are actually very easy to verify


they do not break anything

PR #1290, opened 2018-09-02
travis-ci: supress build log if build was successful

chipitsine:


dlemstra:
@chipitsine Thanks for sending us all these pull requests. And I don't mind that you don't add an explanation to those null pointer dereferences pull requests. But this one really needs an explanation. Why would we want to change this?

chipitsine:
well, there could be build time warnings.
if none of those warnings are interesting, so we can supress (successful) output

dlemstra:
Closing this as we are planning to move to GitHub Actions CI.

PR #1291, opened 2018-09-02
MagickCore/montage.c: resolve null pointer dereference

chipitsine:


dlemstra:
We have applied a slightly different patch: 7c7eb855afb90ad70b7fbe69a37239a0fbfc1648

PR #1308, opened 2018-09-13
coders/pgx.c: remove unused code

chipitsine:
found by coverity

there's already check against NULL in line 176


dlemstra:
Thanks 👍

PR #1311, opened 2018-09-14
MagickCore/pixel-accessor.h: resolve null pointer dereference

chipitsine:
we need to put second "if" inside first "if" to avoid possible null pointer dereference


dlemstra:
Good catch. Could you please reformat the code so it has the correct indentation?

chipitsine:
A bit later. I will open PR for 6 as well

chipitsine:
I hope that I aligned well now

dlemstra:
Almost there 👍 Line 422 and 423 need some extra spaces.

chipitsine:
added  extra spaces to those lines

PR #1316, opened 2018-09-16
MagickWand/operation.c: resolve possible null pointer dereference

chipitsine:
found by coverity


urban-warrior:
We added a check inside the StringToArrayOfDoubles() method instead.  Thanks for the heads-up.

PR #1317, opened 2018-09-16
MagickCore/fx.c: revolve null pointer dereference

chipitsine:
found by coverity


urban-warrior:
Thanks for alerting us to the problem.  We applied a slightly different patch that what you suggested.

PR #1326, opened 2018-09-22
MagickCore/property.c: resolve null pointer dereference

chipitsine:
found by coverity


urban-warrior:
Thanks for the problem report and patch.  Our patch differs slightly from your suggestion but does resolve the null pointer dereference.

PR #1328, opened 2018-09-23
Magick++/lib/Options.cpp: remove redundant assignment

chipitsine:
found by PVS analyzer


dlemstra:
Could you instead of removing those lines move the assignment to that spot?

chipitsine:
done

PR #1339, opened 2018-10-07
cleanup identical conditions

chipitsine:
[MagickCore/nt-base.c:1403] -> [MagickCore/nt-base.c:1408]: (warning) Identical inner 'if' condition is always true.
[MagickCore/prepress.c:139] -> [MagickCore/prepress.c:144]: (warning) Identical inner 'if' condition is always true.
[MagickCore/quantize.c:3382] -> [MagickCore/quantize.c:3387]: (warning) Identical inner 'if' condition is always true.
[MagickCore/xml-tree.c:449] -> [MagickCore/xml-tree.c:454]: (warning) Identical inner 'if' condition is always true.
[MagickCore/xml-tree.c:528] -> [MagickCore/xml-tree.c:530]: (warning) Identical inner 'if' condition is always true.
[coders/dds.c:1172] -> [coders/dds.c:1178]: (warning) Identical inner 'if' condition is always true.

### DIFF
--- a/MagickCore/nt-base.c
+++ b/MagickCore/nt-base.c
@@ -1405,22 +1405,19 @@ MagickPrivate int NTGhostscriptEXE(char *path,int length)
       if (ghost_semaphore == (SemaphoreInfo *) NULL)
         ActivateSemaphoreInfo(&ghost_semaphore);
       LockSemaphoreInfo(ghost_semaphore);
-      if (*program == '\0')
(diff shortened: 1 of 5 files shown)


urban-warrior:
See https://en.wikipedia.org/wiki/Double-checked_locking.  Most / all of your clean-up modifies the threading logic.  Your patches could introduce incorrect behavior when threaded is enabled.

chipitsine:
thank for clarification.
will have a deeper look on threads

PR #1355, opened 2018-10-17
delegates.xml: added missing quality flag for WebP encode

brusch:
### Description

Currently the `-quality` flag is not considered when target format ist WebP. 
This PR adds the `-q` flag of `cwebp` to the encode delegate of WebP, that solves the problem.


dlemstra:
We also have the same issue here: https://github.com/ImageMagick/ImageMagick/blob/master/MagickCore/delegate.c#L123.

brusch:
Missed that one, thanks! 
Just updated my PR.

dlemstra:
Thanks for helping us out!

Issue #1266, opened 2018-08-26
Bug in writing EXIF metadata during HEIC to JPEG conversion


asuwend1:
### Description
When I convert HEIC with EXIF metadata to JPEG, I can't read the EXIF metadata properly from the resulting JPEG **with some commonly used Java packages** (exiftool can still read it with warning though). I believe this used to work properly approximately 1 month ago.

Metadata reader tried:
https://github.com/apache/commons-imaging
https://github.com/drewnoakes/metadata-extractor

### Steps to Reproduce
Input image: https://github.com/monostream/tifig/blob/master/fixtures/5_tree_path.heic
(this input image has EXIF Orientation tag)
asuwedi-mbp:tmp Ari$ exiftool 5_tree_path.heic | grep -i warn
asuwedi-mbp:tmp Ari$ convert 5_tree_path.heic 5_tree_path.jpg
asuwedi-mbp:tmp Ari$ exiftool 5_tree_path.jpg | grep -i warn
**Warning                         : [minor] Unknown garbage at start of EXIF segment**

I believe, this "minor" warning caused it to be unreadable to many EXIF metadata reader.

asuwedi-mbp:tmp Ari$ exiftool -v5 5_tree_path.jpg

...
    07f8: 00 00 00 06 00 00 00 01 00 00                   [..........]
  **Warning = [minor] Unknown garbage at start of EXIF segment**
  ExifByteOrder = MM
  + [IFD0 directory with 11 entries]
...

### System Configuration

- ImageMagick version:
Version: ImageMagick 7.0.7-39 Q16 x86_64 2018-08-25 https://www.imagemagick.org
- Environment (Operating system, version and so on):
MacOS High Sierra, 10.13.6
- Additional information:

dlemstra:
Thanks for the problem report. We can reproduce it and will have a patch to fix it in GIT master branch @ https://github.com/ImageMagick/ImageMagick later today. The patch will be available in the beta releases of ImageMagick @ https://www.imagemagick.org/download/beta/ by sometime tomorrow.

PR #1397, opened 2018-11-26
Fix heic exif extraction bug where we are seeing 4 unwanted bytes before EXIF magic number

asuwendi:
### Prerequisites

- [ x ] I have written a descriptive pull-request title
- [ x ] I have verified that there are no overlapping [pull-requests](https://github.com/ImageMagick/ImageMagick/pulls) open
- [ x ] I have verified that I am following the existing coding patterns and practices as demonstrated in the repository.

### Description
This code change would fix: https://github.com/ImageMagick/ImageMagick/issues/1266
              Refering to JpegEncoder::Encode in libheif's encoder_jpeg.cc
              https://github.com/strukturag/libheif/blob/master/examples/encoder_jpeg.cc
              in the call to jpeg_write_marker, we need to manually remove
              the first 4 bytes returned by heif_image_handle_get_metadata
              JFIF APP1 marker should point to the EXIF magic number
              0x45 0x78 0x69 0x66/"Exif"
              If this is not done, some EXIF readers can't read the EXIF metadata

### DIFF
--- a/coders/heic.c
+++ b/coders/heic.c
@@ -290,7 +290,15 @@ static Image *ReadHEICImage(const ImageInfo *image_info,
               StringInfo
                 *profile;
 
-              profile=BlobToStringInfo(exif_buffer,exif_size);
+              //Refering to JpegEncoder::Encode in libheif's encoder_jpeg.cc
+              //https://github.com/strukturag/libheif/blob/master/examples/encoder_jpeg.cc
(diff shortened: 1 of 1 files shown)


dlemstra:
Thanks for your PR and helping us find the issue. We have applied a different version of your patch that includes some extra checks to the git repo. This will be resolved in the next release.

PR #1436, opened 2019-01-03
ignore_transformations when `heic:preserve-orientation=true`

liuzhe0223:
### Prerequisites

### Description

 Previous fix for #1232 is not good enough with photos that has no `heif irot` but only `exif:Orientation`.

The `heic` coder will transform the origin pixels with properties (irot, imir, clap) of heif default.

Example photo in  #1232 has both `heif irot` and `exif:Orientation`.  And the two will rotate origin  pixels with the same degrees. So previous fix got correct result.

This pull-request use `ignore_transformations` decode options  to keep the origin pixels orientation instead of  `CompensateOrientation`.

### Example photo
 https://www.dropbox.com/s/bzkbemy7ngiaghb/FvwrpJFxWoehjaYqtnaXa29TauF8.heic?dl=0
This photo has no `heif irot` but only `exif:Orientation`.

### DIFF
--- a/coders/heic.c
+++ b/coders/heic.c
@@ -162,7 +162,7 @@ static Image *ReadHEICImage(const ImageInfo *image_info,
     *image_handle;
 
   struct heif_decoding_options
-    *decode_options=NULL;
+    *decode_options;
 
   uint8_t
     *p_y,
@@ -173,7 +173,7 @@ static Image *ReadHEICImage(const ImageInfo *image_info,
     *file_data;
 
   const char
-    *preserve_orientation_option;
+    *option;
 
   /*
     Open image file.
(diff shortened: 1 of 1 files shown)


liuzhe0223:
Thank you for your review , I have fixed code according to your comments @dlemstra .

coders/heic.c
```diff
@@ -331,8 +285,21 @@ static Image *ReadHEICImage(const ImageInfo *image_info,
...
   (void) SetImageColorspace(image,YCbCrColorspace,exception);
+  decode_options=(struct heif_decoding_options *) NULL;
+  option=GetImageOption(image_info,"heic:preserve-orientation");
+  if (IsStringTrue(option) == MagickTrue)
+    {
+      decode_options=heif_decoding_options_alloc();
+      decode_options->ignore_transformations=1;
+    }
+  else
+    {
```
dlemstra: Almost there 👍 Thanks for making all these changes for us. 

We don't use braces when the statement is just one line so could you remove them here?

liuzhe0223: ok

liuzhe0223: done

PR #1496, opened 2019-03-05
Honour SOURCE_DATE_EPOCH

jschueller:
Honour the SOURCE_DATE_EPOCH environment variable when writing
to some files: PS, PDF, PNG, CIN, DPX, MAT, PDB, setting the timestamp
to the provided value and using GM time instead of local time.

From: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=819914

### Prerequisites

### DIFF
--- a/MagickCore/string.c
+++ b/MagickCore/string.c
@@ -1156,6 +1156,81 @@ MagickExport ssize_t FormatMagickSize(const MagickSizeType size,
   return(count);
 }
 
+/* CurrentTime()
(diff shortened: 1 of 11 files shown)


urban-warrior:
We're adding support for SOURCE_DATE_EPOCH but are manually adding the patches because we took a slightly different approach.  The work is not complete yet but should be by end-of-week.  We will reference this pull request in the Changelog.

jschueller:
thanks !

PR #1563, opened 2019-04-29
Fix rubber banding artifacts in the XSelectWindow function. #1460 

### LINKED ISSUE #1460: Region selection in "import" command produce corrupted graphics and output (Unexpected lines).
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am using the latest version of ImageMagick
- [x] I have searched [open](https://github.com/ImageMagick/ImageMagick/issues) and [closed](https://github.com/ImageMagick/ImageMagick/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When taking a screenshot using the command `import test.png`, the rectangle lines used to visualize the selected region leave trails and artifacts, those lines are preserved and are visible in the final image. This gif illustrates the problem:

Problem

### Steps to Reproduce
Execute: `import test.png`

### System Configuration

- ImageMagick version: 7.0.8
- Environment (Operating system, version and so on): Arch Linux. Deepin Desktop Environment. Using AMDGPU driver with Mesa's RadeonSI OpenGL implementation on an RX 590.
- Additional information:
```
Features: Cipher DPC HDRI Modules OpenCL OpenMP 
Delegates (built-in): bzlib cairo fontconfig freetype gslib heic jbig jng jp2 jpeg lcms lqr ltdl lzma openexr pangocairo png ps raw rsvg tiff webp wmf x xml zlib
```
This [question](https://superuser.com/questions/1295713/weird-border-artefacts-when-taking-screenshots-with-import) on SuperUser seem to be related.

OmarSquircleArt:
### Prerequisites

### Description

This patch fixed #1460 . The problem was caused by the fact that applications redraw regions in which selection happen, so rubber banding doesn't work. A simple fix is to grab the X server before doing the selection. To quote the *Xlib Programming Manual*:

> XGrabServer() and XUngrabServer() are used when a  program requires total control of the screen, so that output requests from other programs are queued but not displayed. One application of grabbing is to draw temporary moving objects on the screen, such as the outline of a window being moved. This is called rubber−banding.

### DIFF
--- a/MagickCore/xwindow.c
+++ b/MagickCore/xwindow.c
@@ -9315,6 +9315,7 @@ static Window XSelectWindow(Display *display,RectangleInfo *crop_info)
   target_window=(Window) NULL;
   x_offset=0;
   y_offset=0;
+  (void) XGrabServer(display);
   do
   {
     if ((crop_info->width*crop_info->height) >= MinimumCropArea)
@@ -9383,6 +9384,7 @@ static Window XSelectWindow(Display *display,RectangleInfo *crop_info)
         break;
     }
   } while ((target_window == (Window) NULL) || (presses > 0));
+  (void) XUngrabServer(display);
   (void) XUngrabPointer(display,CurrentTime);
   (void) XFreeCursor(display,target_cursor);
   (void) XFreeGC(display,annotate_context);


charbelnicolas:
Can this patch be the reason why the selection rectangle on the import command doesn't respect neither picom or nvidia vsync settings?

I've noticed tearing while resizing the selection rectangle on 7.0.11.13.

urban-warrior:
The import command worked for years without complaint.  In the last few years we started getting reports of streaks across the screen capture from the outline of the rubber band.  Grabbing the server fixed this issue for all the use cases presented to us.  On the surface, grabbing the screen for the exclusive use of rubber-banding while identifying a rectangle to capture on the screen, seems unlikely to be a root cause of failure with picom.  You certainly could remove XGrabServer() from xwindow.c, rebuild, and grab the screen using picom.  Does it work now?  If so, we're clueless how to resolve a screen grab so it satisfies all window managers.

charbelnicolas:
Hi, I'll try to remove it and see if it works! BTW, is there a way to remove the crosshair from the import command? I think it gets in the way of properly selecting an area or is there a setting to remove it? (Sorry if this is not the place to ask)

urban-warrior:
We need a pointer of some type to help identify the boundaries of the screen import.  Do you have a better cursor we should consider other than the cross hairs?

charbelnicolas:
Ok, I compiled it and ran the import command like this:

ImageMagick/utilities > . import.1

And the issue is still there, so I guess that's not it. By the way, is that the correct way to test it? Or is there a more correct way to do it?

charbelnicolas:
> We need a pointer of some type to help identify the boundaries of the screen import. Do you have a better cursor we should consider other than the cross hairs?

I think the cursor is not that bad at the moment but it could be slimmer. A vertically inverted L shape could be better to align the selection rectangle edges. To solve any visibility issues that may arise I'd make the L shape line color inverse to whatever color is behind it in the background.

Maybe something like this? (Not trying to reinvent the wheel)

charbelnicolas:
I did the sudo make install and tried it without the XGrabServer() and it made the selection rectangle behave and display weirdly, so that definitely was not it.

PR #1571, opened 2019-05-15
Implementation of a TIM2 coder

Qjammer:
### Prerequisites

### Description
This pull request implements a coder for the TIM2 format used in PS2 videogames.

It can handle two format types (4.0, 4.1), three pixel depths (RGBA16, RGB24, RGBA32) with or without a CLUT (Color LookUp Table) of either 4 or 8 bit indices in either of two CLUT Storage Modes (CSM1, CSM2).

It does not handle mipmaps or multiple image files, as I haven't found an example to inspect.


Qjammer:
The build fails because automake needs to be run, since a new file has been added. I don't understand why autogenerated files are commited in the git history, like the utilities/*.1 files, which are generated by `$ ./configure` .

I think the CI build should include the command `$ automake` to generate a new Makefile.in file, and then run `$ configure` as it does now.

I am therefore unsure if I should add a commit changing the autogenerated files, in particular, the Makefile.in

PR #1661, opened 2019-08-08
Added support for DWAA/B EXR compression methods

flablog:
### Prerequisites

### Description
**Problem**
- It was impossible to write an EXR with the DWAA/DWAB Compression methods even though everything is present in the Exr library

**Solution**
- Add the missing defines/includes


MagickCore/compress.h
```diff
@@ -43,6 +43,8 @@ typedef enum
   NoCompression,
   PizCompression,
   Pxr24Compression,
+  DwaaCompression,
```
dlemstra: These values should be below `WebPCompression` as the last two values. This should be done because we don't want to change the numeric value of the current values.

Magick++/lib/Magick++/Include.h
```diff
@@ -518,6 +518,8 @@ namespace Magick
   using MagickCore::LZWCompression;
   using MagickCore::PizCompression;
   using MagickCore::Pxr24Compression;
+  using MagickCore::DwaaCompression;
```
dlemstra: These values are ordered alphabetically. Could you move the two new values?

MagickCore/option.c
```diff
@@ -1241,6 +1241,8 @@ static const OptionInfo
     { "None", NoCompression, UndefinedOptionFlag, MagickFalse },
     { "Piz", PizCompression, UndefinedOptionFlag, MagickFalse },
     { "Pxr24", Pxr24Compression, UndefinedOptionFlag, MagickFalse },
+    { "Dwaa", DwaaCompression, UndefinedOptionFlag, MagickFalse },
```
dlemstra: These values are ordered alphabetically. Could you move the two new values?

coders/exr.c
```diff
@@ -453,6 +453,10 @@ static MagickBooleanType WriteEXRImage(const ImageInfo *image_info,Image *image,
     compression=IMF_PIZ_COMPRESSION;
   if (write_info->compression == Pxr24Compression)
     compression=IMF_PXR24_COMPRESSION;
+  if (write_info->compression == DwaaCompression)
```
dlemstra: Because these values are not yet available you cannot just use them here. For these two values you will need to check if the compression value is defined:

```c
#if defined(IMF_DWAA_COMPRESSION)
  if (write_info->compression == DwaaCompression)
    compression=IMF_DWAA_COMPRESSION;
#endif
```

Could you add these two new values below `IMF_B44A_COMPRESSION`? You might notice that the check is different but that is because I just pushed a patch to fix this. And could you also add support for reading these two new values in `ReadEXRImage`? I just pushed some another patch to add support for reading the compression.

flablog: Thanks for describing in details what I needed to update! I've made an update to the merge request considering all your comments.

flablog:
Thanks for describing in details what I needed to update! I've made an update to the merge request considering all your comments.

dlemstra:
Thanks for the patches. Would you mind also implementing the part inside `ReadEXRImage`?

Magick++/lib/Magick++/Include.h
```diff
@@ -517,7 +519,7 @@ namespace Magick
   using MagickCore::LZMACompression;
   using MagickCore::LZWCompression;
   using MagickCore::PizCompression;
-  using MagickCore::Pxr24Compression;
+  using MagickCore::Pxr24Compression;
```
dlemstra: Could you please remove the extra space?

flablog: Oups, I'll clean that for sure

For the ReadEXRImage, I'll check that too

flablog: What is weird is that convert is completly capable of reading Dwaa exr without the lines I'm adding

flablog: (The extra space have been fixed in an amend commit)

dlemstra: What that part of the code does it setting the compression on the image. It does not need it to be able to read the file. You can see that with when you use identify:

```
I:\exr>identify -format "Compression: %[compression]\n" *.exr
Compression: Zip
Compression: None
Compression: ZipS
Compression: ZipS
Compression: Piz
Compression: Piz
Compression: Piz
Compression: Piz
Compression: Piz
```

dlemstra:
Thanks for making the changes in the read method. I am sorry but I would like to request another change. Could you change the casing of `Dwaa` to `DWAA` and `Dwab` to `DWAB`?

flablog:
In the coders/exr.c ? You mean here for example ?
```
#if defined(IMF_DWAA_COMPRESSION)
  if (compression == IMF_DWAA_COMPRESSION)
    image->compression=DWAACompression;
#endif
```
I wasn't shure when to use upercases. B44A and RLE for example uses upercases, but Piz, Zip, Pxr24 are not.
Should I fix them too ?

dlemstra:
I did mean everywhere. But you cannot change the other values because that would break it for people that are using our code.

flablog:
Done. Thanks for taking so much time to explain et correct me !

Issue #1749, opened 2019-10-13
outside the range of representable values of type 'unsigned int' at coders/bmp.c


SuhwanSong:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am using the latest version of ImageMagick
- [x] I have searched [open](https://github.com/ImageMagick/ImageMagick/issues) and [closed](https://github.com/ImageMagick/ImageMagick/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

There are 9 outside the range of representable values of type 'unsigned int' at coders/bmp.c
### Steps to Reproduce

[poc](https://drive.google.com/file/d/12wejXA-e4IJ9xZggBWmICsZbHOFgzDHN/view?usp=sharing)
please run a following cmd with poc file.
`magick  -blue-primary 210,448 -red-primary 146,468 $PoC -depth 59 -treedepth 91 -green-primary 850,945 tmp.rla`

Here's UBSAN log
```
coders/bmp.c:2307:39: runtime error: 1.56766e+11 is outside the range of representable values of type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior coders/bmp.c:2307:39 in 
coders/bmp.c:2309:39: runtime error: 5.02511e+11 is outside the range of representable values of type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior coders/bmp.c:2309:39 in 
coders/bmp.c:2311:39: runtime error: -6.58204e+11 is outside the range of representable values of type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior coders/bmp.c:2311:39 in 
coders/bmp.c:2314:39: runtime error: 9.12681e+11 is outside the range of representable values of type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior coders/bmp.c:2314:39 in 
coders/bmp.c:2316:39: runtime error: 1.01469e+12 is outside the range of representable values of type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior coders/bmp.c:2316:39 in 
coders/bmp.c:2318:39: runtime error: -1.92629e+12 is outside the range of representable values of type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior coders/bmp.c:2318:39 in 
[...]

PR #1853, opened 2020-02-27
Fix 1749

g0483hqg3:
### Prerequisites

### Description
The change for #1749 ([link](https://github.com/ImageMagick/ImageMagick/commit/a1142af44f61c038ad3eccc099c5b9548b507846?diff=split)) breaks the functionality because the cast to ssize_t _always_ results in 0. This requests does bounds checking instead and sets the value to .0 if out-of-bounds.

### DIFF
--- a/Magick++/fuzz/encoder_fuzzer.cc
+++ b/Magick++/fuzz/encoder_fuzzer.cc
@@ -33,7 +33,8 @@ static ssize_t EncoderInitializer(const uint8_t *Data, const size_t Size, Magick
   return 0;
 }
 
-extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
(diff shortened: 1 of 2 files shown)


Magick++/fuzz/encoder_fuzzer.cc
```diff
@@ -33,8 +33,7 @@ static ssize_t EncoderInitializer(const uint8_t *Data, const size_t Size, Magick
   return 0;
 }
 
-extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
-{
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
```
dlemstra: Looks like this got into your PR by accident?

g0483hqg3: Indeed.

dlemstra:
Thanks 👍

PR #2195, opened 2020-06-24
Issue 85

remicollet:
Same as https://github.com/ImageMagick/ImageMagick6/pull/86


urban-warrior:
Your pull broke a number of unit tests.  We will need to revert.

remicollet:
This seems very strange as for me both ImageMagick6 and imagick extension test suite passes with no failure....

remicollet:
To summarize my thinking

* I don't think that `heigh == 0` is bad, as this is expected when `interline` is explicitly set to 0
* I think the problem is the default value of `interline` being 0

Using `pointsize` when heigh is 0 (current code) seems a bad idea, so I've tried to fix proper default value of `interline = pointsize`.

Notice: my RPM builds for 21 include this PR, so will see if some users come to report issues (already hundreds of downloads, usually ~10k for each version)

/cc @Danack as it seems discussion have moved here.

PR #2231, opened 2020-07-03
Ensure that memory is initialized

g0483hqg3:
### Prerequisites

### Description
Currently, the allocation method relies on the fact that all its callers correctly initialize the memory. This is an error-prone pattern and this patch will address this problem.

### DIFF
--- a/MagickCore/memory.c
+++ b/MagickCore/memory.c
@@ -338,12 +338,18 @@ MagickExport void *AcquireAlignedMemory(const size_t count,const size_t quantum)
 {
   size_t
     size;
+  void
+    *memory=NULL;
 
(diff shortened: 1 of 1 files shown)


urban-warrior:
We do not zero memory by design.  In many cases, memory is set by the algorithm making zeroing an unnecessary reduntant performance hit.  You can allocate the pixel cache and some internal buffers with anonymous memory mapping rather than from heap with this security policy. As a consequence, the pixels are initialized to zero resulting in a minor performance penalty.

```
<policy domain="system" name="memory-map" value="anonymous"/>
<policy domain="cache" name="memory-map" value="anonymous"/>
```
Otherwise, you can zero memory with the ResetMagickMemory() method.

g0483hqg3:
Ok, makes sense. Thanks!

PR #2275, opened 2020-07-15
Fix -deconstruct replacement

stewbasic:
### Prerequisites

### Description
Fix -deconstruct replacement
Change from -layer to -layers, as the former is unrecognized and the latter is consistent with www/porting.html#L575.


dlemstra:
Thanks for fixing this for us 👍

PR #2327, opened 2020-07-27
PNG-Encoder: Fix missing exception inheritance.

grindhold:
a call to InheritException was missing, leading to cryptic error
messages in case of cache exhaustion. now the cache exhaustion
error gets propagated to the user properly

### Prerequisites

### Description
We need to use InheritException in the same fashion we do in JPG or TIFF, in order to have beautiful error messages at the user-side.


dlemstra:
This should fail on all platforms because the image has no `exception` field. We will need to investigate why this is happening. Did you plan to apply this patch to IM6?

grindhold:
indeed, i found this bug in 6.9

grindhold:
nice

Issue #1547, opened 2019-04-15
Support for Krita KRA and ORA images format


cgilles:
Krita image editor as a documented multi-layer and compressed format to store edited image : KRA.

From Krita documentation: ".kra is Krita’s internal file-format, which means that it is the file format that saves all of the features Krita can handle. It’s construction is vaguely based on the open document standard, which means that you can rename your .kra file to a .zip file and open it up to look at the insides."

As for Gimp XCF, ImageMagick must be able at least to extract layers from this container.

Source code of KRA and ORA to QImage thumbnailers : 

https://cgit.kde.org/kimageformats.git/tree/src/imageformats/kra.cpp
https://cgit.kde.org/kimageformats.git/tree/src/imageformats/ora.cpp

In fact, the KRA container is a zip like compressed tree with all layers stored with PNG data. ORA sound like very similar.

SethX3:
Very interested in OpenRaster (.ora) as a potential open alternative to PSD so would like this very much.

Christopher-Chianelli:
Update on this issue: OpenRaster 0.2 and above decoder support is merged. It only extracts the mergedimage.png from the file, so extracting layers is not yet supported and is future work that need to be done. I'll look at the Krita format to see how similar it is to OpenRaster.

french-paragon:
.kra files have a mergedimage.png too, so I guess it should work as is. (From what I know thus, the mergedimage do not retain the color profile).

Christopher-Chianelli:
Was able to get documentation on the KRA format from freenode: https://invent.kde.org/documentation/docs-krita-org/-/merge_requests/105/diffs. `mergedimage.png` is optional in Krita (it is not present in auto-saves or backups), but will almost always be present in the "main" Krita file (since 2014).

Christopher-Chianelli:
From the doc: " The format of the pixel data
is internal to Krita and dependent on the version of Krita and can change 
without warning.", which means extracting layer data is basically a no-go and the best we can do is extract `mergedimage.png`, or else `preview.png` if it does not exist (in the case of backups/autosave)

PR #2342, opened 2020-07-31
(DO NOT MERGE) OpenRaster (.ora) decoder support

Christopher-Chianelli:
### Prerequisites

- [ ] I have verified that I am following the existing coding patterns and practices as demonstrated in the repository (I did my best to emulate the code style, but I am unfamiliar with this repository)

### Description
The OpenRaster format (https://www.freedesktop.org/wiki/Specifications/OpenRaster/) is a format used in Krita and MyPaint. As such, it be useful to be able to read from said format to modify any Picture created in said app to, for instance:
- Convert it to a format better supported on the web
- Apply any of ImageMagick many operations to the Picture
- etc.

The OpenRaster format is a ZIP Archive following the format in https://www.freedesktop.org/wiki/Specifications/OpenRaster/Draft/FileLayout/. Of particular note, the archive contains "/mergedimage.png", which contain the final rendered image. Thus, to implement support, I read that file from the ZIP archive.

Fixes half of https://github.com/ImageMagick/ImageMagick/issues/1547, I need to research more about Krita file format before I can create the decoder of .kra

### Risks

Sadly, it doesn't appear any of the current dependencies have ZIP Archive support; the compression libraries I saw from a quick glance of the code only compress/decompress, not read from an archive. As such, I added libzip 1.0.0 as a dependency. 

### TODO

- Tests
- <del>Actually make it work; the code provided is just to 1. get it to build the additional decoder and 2. be a basic outline of how it should work. Currently, it doesn't recognize my decoder to be the decoder for .ora files, and it probably have some bugs since I am unfamiliar with this code base. Decided to open this PR early to get feedback.</del>


Christopher-Chianelli:
Update: figured out why my decoder wasn't being detected; I set it format_type to implict, which meant it doesn't assume type from file name. Changing it to explicit fixes it. I know the image it extracts to the temp file is correct (I am able to view the image on my computer), but despite explicitly setting the image type to be PNG, I get
```
identify: NoDecodeDelegateForThisImageFormat `PNG' @ error/constitute.c/ReadImage/562.
```
it might be because the temporary file I write to have no file extension?

Christopher-Chianelli:
(Look like I don't have PNG image support; probably need to install libpng)

Christopher-Chianelli:
Okay, got everything to work. It should now also degrade to not having ORA support if libzip is not installed.

Christopher-Chianelli:
Future work: add support for extracting layers from the ORA image using https://www.freedesktop.org/wiki/Specifications/OpenRaster/Draft/LayersStack/ (note: ORA image may contain SVG, PNG and others, and you need to do transforms on them to get the mergedimage.png; extracting the mergedimage.png is less prone to errors than combining all the layers together, but doesn't allow extracting individual layers).

urban-warrior:
Use /* */ in your comments.  Not all C compilers accept // as a comment (e.g. // #if defined(MAGICKCORE_LIBZIP_DELEGATE)).

In RegisterORAImage(), do not conditionally compile the entire block.  Instead, conditionally compile only the `entry->decoder...` statement.

Can you post a link to a few reference ORA image files?  We will need those for our image reference library.

Christopher-Chianelli:
@urban-warrior Here are two sample images (Found in the MyPaint test suite): https://github.com/mypaint/mypaint/raw/master/tests/bigimage.ora https://github.com/mypaint/mypaint/raw/master/tests/smallimage.ora

Christopher-Chianelli:
(I believe those images are licensed under GPL, so you may not want to include them directly in the repo. I can also create some OpenRaster images if that be of assistance).

urban-warrior:
Exported an ORA image from gimp and could read it with your patch.  However, for your samples, we get:

```
convert bigimage.ora bigimage.png
convert: unable to open file 'bigimage.ora': No such file or directory @ error/ora.c/ReadORAImage/155.
convert: no images defined `bigimage.png' @ error/convert.c/ConvertImageCommand/3285.
```

urban-warrior:
We won't include the sample ORA images in the ImageMagick repo.  Instead, we have an internal repo we use for offline unit testing of ImageMagick before a release.

Christopher-Chianelli:
I know one of the samples is quite old, perhaps from before when `mergedimage.png` was mandatory (it was optional a long time ago, (before 0.0.2) but became mandatory when it became a more popular format to increase viewer software support). To support those old images, I would need to implement the layering support from reading stack.xml and layering all the images together.

Christopher-Chianelli:
(can confirm it a file from before 0.0.2; no mergedimage.png in root upon unzipping)

Issue #2551, opened 2020-09-11
Converting images AV1 / .avif fails


manniL:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am using the latest version of ImageMagick
- [x] I have searched [open](https://github.com/ImageMagick/ImageMagick/issues) and [closed](https://github.com/ImageMagick/ImageMagick/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

I've tried converting several images of different types (jpg, png, webp) to avif and always receive the following error message:

> magick: Unsupported file-type: Unspecified `icon.avif' @ error/heic.c/IsHeifSuccess/134.

Output of `convert -verbose icon.png me.avif`:

```
icon.png PNG 512x512 512x512+0+0 8-bit sRGB 390912B 0.010u 0:00.014
icon.png=>me.avif PNG 512x512 512x512+0+0 8-bit sRGB 0.010u 0:00.008
convert: Unsupported file-type: Unspecified `me.avif' @ error/heic.c/IsHeifSuccess/134.
```

### System Configuration

- ImageMagick version: Version: ImageMagick 7.0.10-29 Q16 x86_64 2020-09-06
- Environment (Operating system, version and so on): macOS Catalina 10.15.6 19G73 x86_64
- Additional information: libheif: stable 1.8.0 (bottled), aom: stable 2.0.0 (bottled)

dlemstra:
Are you sure that your `libheif` library is build with support for `aom`? You will get this message when the avif encoder could not be found.

manniL:
> Are you sure that your `libheif` library is build with support for `aom`? You will get this message when the avif encoder could not be found.

You are right, I've re-installed `libheif` with `--verbose` and it seems it doesn't detect the lib though both, `rav1e` *and* `aom` are provided through homebrew. 🤔

```
configure: libaom decoder: no
configure: libaom encoder: no
configure: rav1e encoder: no
```

Not an ImageMagick issue so I'm closing here. I'm open for ideas how to solve this though 😋

Issue #1432, opened 2018-12-28
Add AVIF (AV1 Image File Format) support


EwoutH:
AVIF (AV1 Image File Format) is a new image format based on the AV1 video codec. It supports both lossless and lossy compression and [saves](https://calendar.perfplanet.com/2018/is-avif-the-future-of-images-on-the-web/) up to 50% space at the same quality compared to JPG.

Here is the [AVIF specification](https://aomediacodec.github.io/av1-avif/) and some .avif samples from [Netflix](http://download.opencontent.netflix.com/?prefix=AV1/Chimera/AVIF/) and [Microsoft](https://github.com/AOMediaCodec/av1-avif/tree/master/testFiles/Microsoft). AVIF files can currently be viewed with [VLC](https://nightlies.videolan.org/) and are going to be supported soon in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1443863) and [Windows 10](https://github.com/AOMediaCodec/av1-avif/wiki).

EwoutH:
[libavif](https://github.com/joedrago/avif) tagged it's first release, [0.1.0](https://github.com/joedrago/avif/releases/tag/v0.1.0).

EwoutH:
[libavif](https://github.com/joedrago/avif) tagged a new release, [0.1.3](https://github.com/joedrago/avif/releases/tag/v0.1.3). See [Colorist](https://github.com/joedrago/colorist) for an example [implementation](https://github.com/joedrago/colorist/blob/master/lib/src/format_avif.c).

novomesk:
From http://www.imagemagick.org/discourse-server/viewtopic.php?p=172532&sid=29c77b7c1d3fe14155ad1a3d255707b6#p172532 comment, ImageMagick devs would prefer using libheif for AVIF support. Unfortunately the AVIF support stays in experimental branch only and the progress stopped.
I asked [What is the future of AVIF support in libheif?](https://github.com/strukturag/libheif/issues/217) but so far there is no reply.

In order to make libheif to read/write AVIF, there is still need of another dependency: [libaom](https://aomedia.googlesource.com/aom/) - to encode/decode AV1 streams.

[...]

PR #2733, opened 2020-10-17
Fix AVIF support reporting

siiptuo:
# Prerequisites

### Description

As in #1432 and #2551 libheif may not be built with AVIF support but `magick -list format` always reports that reading and writing is supported. This pull request will actually check that libheif supports AVIF.

Previously running `magick input.jpeg output.avif` did give a somewhat nice "Unsupported file-type" message but this pull request will silently output a JPEG. This seems to be consistent with behavior of unknown file extensions but hopefully doesn't cause additional confusion.


dlemstra:
Thanks for the patch. I have mixed feeling about if but after discussing this internally we decided to apply your patch.

PR #2823, opened 2020-11-13
svg: do not pass background to inkscape with opacity=0

zx2c4:
If 0 opacity is passed to Inkscape at the same time as a background
color, the background color wins, and the opacity is actually set to
100%. Fix this by passing the empty string as a background color for
opacity=0.


urban-warrior:
This is a bug in inkscape.  The inkscape developers are aware of the bug and are working on a patch.  In the mean-time, we have a workaround that support newer and older versions of inkscape.  We'll export a patch likely by tomorrow.

zx2c4:
Thanks! I'll close this PR then.

PR #2832, opened 2020-11-16
RFC: stabilize ABI on s390x by tieing MagickFloatType to double

mhillenbrand:
### Prerequisites

### Description

tl;dr: On s390x, `float_t` is defined as `double` for no good reason, yet with unexpected and unnecessary impact on performance in some scenarios. For cleaning that up, `float_t` will have to become `float` on s390x. That would break the ABI of ImageMagick for HDRI=1 on s390x for existing binaries. To avoid that breakage, I propose to set `MagickFloatType` to always be double on s390x so that the API/ABI for HDRI=1 remains as it is today, independent of a change in `float_t`.

What do you think about that approach? Any alternatives I may have missed?

### Background

`float_t` is meant to represent the range and precision that `float` expressions are evaluated in internally. While the s390x architecture very well supports operations on single-precision `float` and gcc by default uses these single-precision operations for that purpose, on s390x `float_t` has historically been defined as `double`. The reason is that the port of glibc for s390x picked up the then-default of setting `float_t` to double.

That combination of gcc's default behavior and `float_t` being `double` contradicts the C standard. To rectify that, gcc _mimics_ behavior according to the standard by evaluating in `double` when called in standard-compliant mode (e.g., `std=c99`): it explicitly converts the values of `float` variables to `double` in registers, performs calculations in double precision, and then rounds and truncates them back to `float` -- at the cost of overhead for the conversions.

To avoid that overhead and be compliant with the C standard (and common sense), we are discussing plans to change `float_t` to `float` and s390 to match gcc's default behavior.

### Impact on ImageMagick

ImageMagick captures the `sizeof(float_t)` at compile-time of the library in `magick-baseconfig.h`. Headers then define `MagickFloatType` to match. For 8-bit and 16-bit depths and HDRI=1, that `MagickFloatType` is used as the type for pixels's color channels (i.e., `Quantum`).

Changes to `float_t` thus would result in a different ABI. Once ImageMagick is rebuilt with HDRI=1 and an updated gcc/glibc, Quantum would then become 32-bit float. Existing binaries and libraries would expect 64-bit double and thus cease to work with newly built version of the library.

To avoid anyone having a bad day from that incompatibility, we could make the ABI on s390x independent of `float_t` and turn MagickFloatType to always remain `double`.

### DIFF
--- a/MagickCore/magick-type.h
+++ b/MagickCore/magick-type.h
@@ -40,6 +40,7 @@ extern "C" {
 #  define MagickULLConstant(c)  ((MagickSizeType) (c ## ULL))
 #endif
 
+#if !defined(__s390__)
 #if MAGICKCORE_SIZEOF_FLOAT_T == 0
 typedef float MagickFloatType;
 #elif (MAGICKCORE_SIZEOF_FLOAT_T == MAGICKCORE_SIZEOF_FLOAT)
@@ -50,6 +51,9 @@ typedef double MagickFloatType;
 typedef double MagickFloatType;
 #else
 #error Your MagickFloatType type is neither a float, nor a double, nor a long double
+#endif
+#else
+typedef double MagickFloatType;
 #endif
 #if MAGICKCORE_SIZEOF_DOUBLE_T == 0
 typedef double MagickDoubleType;


mhillenbrand:
Thanks for merging.

dlemstra:
Thanks for your contribution and research!

bastien-roucaries:
I think this patch should be dropped at next so dump for imagemagic @urban-warrior could you note that next so bump should drop this ugly patch

urban-warrior:
We can consider reverting this patch and bumping the SO, but we need to know the logic behind the request.  It apparently solved a problem on the s390.  Did it cause headaches for the Debian build?  Do you have a better solution to deal the issue posted here?

bastien-roucaries:
@urban-warrior It solved an ABI breakage on glibc...

On the debian side we will ask for a rebuild. It is not a problem and we will bump the so (sort of). But it is a nice patch for other distrib

That  I means is, next time you bump the so, you MUST revert this patch and use float_t == float even on S390.

Bastien

bastien-roucaries:
I means that if in the future you break the ABI by doing a SO dump, revert this patch. float_t ==float is an improvement on s390

urban-warrior:
We'll look toward reverting in the next point release with the requisite SO bump.

PR #3027, opened 2020-12-24
Improve webp encode

tomwei7:
### Prerequisites

### Description

I find some issues when use ImageMagick convert gif to webp. 

GIF has any profiles data will cause ImageMagick output a empty file, after read some code i figure out ImageMagick try to set image profiles after call `WebPEncode`  https://github.com/ImageMagick/ImageMagick/blob/d5daf30364814a002d625ca62eeb7f84ba9535e1/coders/webp.c#L1074 that will cause `WebPMuxSetImage` return uncatched error. https://github.com/ImageMagick/ImageMagick/blob/d5daf30364814a002d625ca62eeb7f84ba9535e1/coders/webp.c#L1200

gif-with-icc
convert this GIF to webp will output empty file and without any error.

Convert some GIF to webp will use a lot of memory, for example this GIF use 18GB memory to convert to webp.
big-gif
most reasons because when WriteAnimatedWEBPImage will alloc memory for every frame and release memory after everything done, but actual `WebPAnimEncoderAdd ` will copy all pixel into self we can safe release memory after call `WebPAnimEncoderAdd`, we can save half memory usage.

https://github.com/ImageMagick/ImageMagick/blob/d5daf30364814a002d625ca62eeb7f84ba9535e1/coders/webp.c#L883-L899

### DIFF
--- a/coders/webp.c
+++ b/coders/webp.c
@@ -765,9 +765,42 @@ typedef struct PictureMemory {
   struct PictureMemory *next;
 } PictureMemory;
 
-static MagickBooleanType WriteSingleWEBPImage(const ImageInfo *image_info,
-  Image *image,WebPPicture *picture,PictureMemory *picture_memory,
-  ExceptionInfo *exception)
+static const char * WebPErrorCodeMessage(WebPEncodingError error_code)
+{
+  switch (error_code)
+  {
+    case VP8_ENC_OK:
+      return NULL;
+    case VP8_ENC_ERROR_OUT_OF_MEMORY:
+      return "out of memory";
+    case VP8_ENC_ERROR_BITSTREAM_OUT_OF_MEMORY:
+      return "bitstream out of memory";
+    case VP8_ENC_ERROR_NULL_PARAMETER:
+      return "NULL parameter";
+    case VP8_ENC_ERROR_INVALID_CONFIGURATION:
+      return "invalid configuration";
+    case VP8_ENC_ERROR_BAD_DIMENSION:
+      return "bad dimension";
+    case VP8_ENC_ERROR_PARTITION0_OVERFLOW:
+      return "partition 0 overflow (> 512K)";
+    case VP8_ENC_ERROR_PARTITION_OVERFLOW:
+      return "partition overflow (> 16M)";
+    case VP8_ENC_ERROR_BAD_WRITE:
+      return "bad write";
+    case VP8_ENC_ERROR_FILE_TOO_BIG:
+      return "file too big (> 4GB)";
+#if WEBP_ENCODER_ABI_VERSION >= 0x0100
+    case VP8_ENC_ERROR_USER_ABORT:
+      return "user abort";
+    case VP8_ENC_ERROR_LAST:
+      return "error last";
+#endif
+  }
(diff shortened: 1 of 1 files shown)


coders/webp.c
```diff
@@ -862,53 +921,107 @@ static MagickBooleanType WriteAnimatedWEBPImage(const ImageInfo *image_info,
...
     enc_options.verbose = 1;
-
-  image=CoalesceImages(image, exception);
-  first_image=image;
   enc=WebPAnimEncoderNew((int) image->page.width,(int) image->page.height,
     &enc_options);
 
-  head=(PictureMemory *) calloc(sizeof(*head),1);
-  current=head;
+  first_image=CoalesceImages(image,exception);
```
dlemstra, line 933: The reason that image was assigned is because `ThrowWriterException` will call `DestroyImage`. It would probably be a good idea to have a single variable for the `while` loop. Or to use `ThrowBinaryException` and `break` from the loop. This way we we also call `WebPAnimEncoderDelete` that was already a bug in the code before you made your patches.

coders/webp.c
```diff
@@ -862,53 +921,107 @@ static MagickBooleanType WriteAnimatedWEBPImage(const ImageInfo *image_info,
...
   {
     if (WebPPictureInit(&picture) == 0)
       ThrowWriterException(ResourceLimitError,"UnableToEncodeImageFile");
 
-    WriteSingleWEBPImage(image_info, image, &picture, current, exception);
+    WriteSingleWEBPPicture(image_info,image,configure,&picture,exception);
     WebPAnimEncoderAdd(enc,&picture,(int) frame_timestamp,configure);
+    WebPPictureFree(&picture);
+
+    if (WebPAnimEncoderGetError(enc) && strlen(WebPAnimEncoderGetError(enc)))
```
dlemstra, line 945: Do we really need to call `strlen(WebPAnimEncoderGetError(enc)` and I am wondering if we could use the result of the call to `WebPAnimEncoderAdd` to determine if adding the frame failed.

coders/webp.c
```diff
@@ -862,53 +921,107 @@ static MagickBooleanType WriteAnimatedWEBPImage(const ImageInfo *image_info,
...
-  writer_info->mem=(unsigned char *) webp_data.bytes;
+  WebPAnimEncoderAssemble(enc,webp_data);
+
   WebPAnimEncoderDelete(enc);
   DestroyImageList(first_image);
-  FreePictureMemoryList(head);
   return(MagickTrue);
 }
+
+static MagickBooleanType WriteWEBPImageProfile(Image *image,WebPData *webp_data,ExceptionInfo *exception)
```
dlemstra, line 964: Good idea to move this inside the `ifdef`.

dlemstra (commented):
Thanks for helping us out. Got a couple remarks about your changes.

dlemstra:
Also looks like we no longer need the `PictureMemory` after this PR.

coders/webp.c
tomwei7: Thinks for review, i change `ThrowWriterException ` to `ThrowMagickException ` and fix some uncatch error from libwebp.

PR #3038, opened 2020-12-26
-pointsize also affects fonts

jidanni:
In fact this
```
For other settings that affect fonts, see the options
```
line appears about five times on the page... and should really be a template...
Indeed the options mentioned seem to vary,
Also I am using GitHub.com to edit this large file, so it was only easy to fix the first one...

P.S., **after** submitting this pull request I noticed within `<!-- -->`
```
If you want to change something in the 'www' or 'ImageMagick' folder please
     open an issue here instead: https://github.com/ImageMagick/Website
```
Well it is too late for that.


dlemstra:
This pull request should be created here instead: https://github.com/ImageMagick/Website.

PR #3689, opened 2021-05-17
Add mingw-w64 CI using msys2

jeroen:
Adds CI for mingw-w64 using the awesome [msys2 action](https://github.com/msys2/setup-msys2). I adapted the build configuration and patches from upstream msys2 [mingw-w64-imagemagick](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-imagemagick).

I think this may be useful to debug some issues and perhaps fix some compiler warnings.

Update: I commented out `make check` because it takes very long.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,3 +113,29 @@ jobs:
         call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\Tools\VsDevCmd.bat"
         cd %GITHUB_WORKSPACE%\ImageMagick-Windows\VisualMagick
         msbuild VisualStaticMTD.sln /m /t:Rebuild /p:Configuration=Release,Platform=x64
+
+  build_msys2:
+    name: Build MSYS2
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        include: [
+          { msystem: ucrt64, toolchain: ucrt-x86_64 },
+          { msystem: mingw64, toolchain: x86_64 },
+          { msystem: mingw32, toolchain: i686 }
+        ]
+      fail-fast: false
+    steps:
+      - name: Prepare git
+        run: git config --global core.autocrlf false
+      - uses: actions/checkout@v2
+      - uses: msys2/setup-msys2@v2
+        with:
+          install: mingw-w64-${{ matrix.toolchain }}-toolchain base-devel binutils
+          update: true
+      - name: Building ImageMagick
+        run: cd .github/workflows && makepkg-mingw --noconfirm --syncdeps
+        env:
+          MINGW_ARCH: ${{ matrix.msystem }}
+          PKGEXT: ".pkg.tar.xz"
+        shell: msys2 {0}
--- a/.github/workflows/PKGBUILD
+++ b/.github/workflows/PKGBUILD
@@ -1,0 +1,141 @@
+# Adapted from upstream:
+# https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-imagemagick
+
+_realname=imagemagick
+pkgbase="mingw-w64-${_realname}"
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=7.9000
+pkgrel=1
+pkgdesc="An image viewing/manipulation program (mingw-w64)"
+arch=('any')
+mingw_arch=('mingw32' 'mingw64' 'ucrt64' 'clang64')
+url="https://www.imagemagick.org/"
+license=("custom")
+makedepends=("${MINGW_PACKAGE_PREFIX}-cairo"
+             "${MINGW_PACKAGE_PREFIX}-gcc"
+             "${MINGW_PACKAGE_PREFIX}-ghostscript"
+             "${MINGW_PACKAGE_PREFIX}-graphviz"
+             "${MINGW_PACKAGE_PREFIX}-librsvg"
+             "${MINGW_PACKAGE_PREFIX}-libxml2"
+             "${MINGW_PACKAGE_PREFIX}-openexr"
+             "${MINGW_PACKAGE_PREFIX}-pango"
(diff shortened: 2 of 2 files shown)


MagickCore/nt-base.c
```diff
@@ -1800,7 +1800,14 @@ MagickPrivate DIR *NTOpenDirectory(const char *path)
     MagickPathExtent);
   if (length == 0)
     return((DIR *) NULL);
+#ifdef __MINGW32__
+  wchar_t WCDirectorySeparator[strlen(DirectorySeparator)+1];
```
dlemstra, line 1804: This indeed incorrect but the solution is actually to remove appending the backslash. This is written in the MSDN documentation:

> This parameter should not be NULL, an invalid string (for example, an empty string or a string that is missing the terminating null character), or end in a trailing backslash (\).

I just pushed a patch to resolve this.

MagickCore/nt-base.h
```diff
@@ -29,7 +29,11 @@ extern "C" {
 
 #define WIN32_LEAN_AND_MEAN
 #define VC_EXTRALEAN
+#ifndef __MINGW32__
 #define _CRT_SECURE_NO_DEPRECATE  1
```
dlemstra, line 33: Why is this added?

MagickCore/nt-base.h
```diff
@@ -99,7 +103,9 @@ extern "C" {
 #if !defined(closedir)
 #  define closedir(directory)  NTCloseDirectory(directory)
 #endif
-#define MAGICKCORE_HAVE_ERF
+#if !defined(MAGICKCORE_HAVE_ERF)
```
dlemstra, line 106: Why do we need this?

dlemstra (commented):
Thanks for your contribution. It is really a good idea to also test the mingw compilation. I left some comments/questions.

.github/workflows/mingw-w64-imagemagick/PKGBUILD
```diff
@@ -0,0 +1,141 @@
+# Adapted from upstream:
```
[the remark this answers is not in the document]
jeroen: I tried to keep the upstream msys2 naming convention, you could imagine having multiple build flavours each with their own PKGBUILD. But I suppose we could move it into the root if you prefer.

.github/workflows/mingw-w64-imagemagick/PKGBUILD
```diff
@@ -0,0 +1,141 @@
...
+options=('staticlibs' 'strip' '!debug' 'libtool')
+source_dir="$GITHUB_WORKSPACE"
+validpgpkeys=('D8272EF51DA223E4D05B466989AB63D48277377A')
+
+prepare() {
+  cd ${source_dir}
+  # autoreconf -fi
+}
+
+build() {
```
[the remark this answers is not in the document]
jeroen: I think it is best to stay close to the upstream PKGBUILD: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-imagemagick/PKGBUILD

That way we test the entire process and dependencies automatically get resolved, and so on.

MagickCore/distribute-cache.c
```diff
@@ -51,6 +51,10 @@
 /*
   Include declarations.
 */
+#if defined(MAGICKCORE_WINDOWS_SUPPORT)
+#include <winsock2.h>
+#include <ws2tcpip.h>
```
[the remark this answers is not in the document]
jeroen: I have copied this from msys2, but perhaps it is no longer needed: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-imagemagick/002-7.0.4.1-build-fixes.patch

Generally, mingw-w64 is more convervative with which headers get included by default, so it could be that this is actually needed. You could try removing it and see if it still builds...

MagickCore/nt-base.h
jeroen: Otherwise you get million "_CRT_SECURE_NO_DEPRECATE redefined" warnings in the build. I suppose mingw-w64 defines this by default

MagickCore/nt-base.h
jeroen: Same as above, it is defined by default in mingw-w64 (perhaps via configure)?

jeroen:
Thanks for your comments. Feel free to push changes into this branch to test what happens in CI.

MagickCore/nt-base.h
dlemstra: I have applied these changes with a separate commit.

MagickCore/nt-base.h
dlemstra: I have applied these changes with a separate commit.

jeroen:
I have rebased the PR and incorporated your suggestions

.github/workflows/main.yml
```diff
@@ -113,3 +113,28 @@ jobs:
...
+      matrix:
+        include: [
+          { msystem: ucrt64, toolchain: ucrt-x86_64 },
+          { msystem: mingw64, toolchain: x86_64 },
+          { msystem: mingw32, toolchain: i686 }
+        ]
+      fail-fast: false
+    steps:
+      - name: Prepare git
+        run: git config --global core.autocrlf false
```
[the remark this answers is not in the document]
jeroen: Git automatically converts line endings to Windows format, but the PKGBUILD file should remain in unix format.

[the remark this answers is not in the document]
jeroen: An alternative solution would be to set this in .gitattributes but then it might also affect other systems.

yg8ijvjvjv:
Why would ImageMagick add a copyright infringement of MinGW-w64?

dlemstra:
Please revert the commit to store the artifacts @jeroen, we don't want to create artifacts every commit.

And I don't understand what you are trying to tell us @yg8ijvjvjv

jeroen:
@dlemstra ok I have removed that and rebased again

PR #3858, opened 2021-06-23
fix memory corruption in ConcatenateStringInfo

remicollet:
See https://github.com/ImageMagick/ImageMagick6/issues/159


MagickCore/string.c
```diff
@@ -540,7 +539,8 @@ MagickExport void ConcatenateStringInfo(StringInfo *string_info,
       sizeof(*string_info->datum));
   if (string_info->datum == (unsigned char *) NULL)
     ThrowFatalException(ResourceLimitFatalError,"MemoryAllocationFailed");
-  (void) memcpy(string_info->datum+length,source->datum,source->length);
```
remicollet, line 543: length is 'total" length, affected too soon, so we need to copy at "old" length, and set it to new value later

dlemstra:
Thanks for the fix.

PR #4064, opened 2021-08-11
Set color encoding in jxl encoder

deymo:
libjxl requires the color encoding to be set before being able to encode
the pixels. If an ICC is available this patch sets the ICC profile on
the encoded image, otherwise it assumes sRGB.

In the future other image metadata like EXIF should be added here as
well.

Tested encoding a PNG with an ICC profile with `convert` and decoding it
back with `djxl`, colors look the same afterwards.

### DIFF
--- a/coders/jxl.c
+++ b/coders/jxl.c
@@ -519,6 +519,60 @@ ModuleExport void UnregisterJXLImage(void)
 %    o image:  The image.
 %
 */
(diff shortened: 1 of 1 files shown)


dlemstra:
I did not realize that without this patch version 0.5 will no longer be able to encode the image. I applied a patch that is similar to this pull request. This will be resolved in the next release of ImageMagick

PR #4103, opened 2021-08-23
Support option -depth for heic encoder

indlin:
The main advantage of heic format is support for 10 and 12 bit color. libheif support it. So, I added support for -depth flag for heic encoder.

Example:
magick convert  file.tif -depth 10 file.heic

Note:
It's make sense for 16-bit source
For now the libheif (and linux viewers) support only 8, 10 and 12 bits. For wrong bit-depth libheif generate error (convert: Encoder plugin generated an error: Unsupported bit depth: Bit depth not supported by x265)
Looks like Mac os support only 10-bit.

### DIFF
--- a/coders/heic.c
+++ b/coders/heic.c
@@ -861,6 +861,7 @@ static MagickBooleanType WriteHEICImageYCbCr(Image *image,
 
   status=MagickTrue;
   bit_depth=8;
+  if (image->image_info->depth > 0) bit_depth=image->image_info->depth;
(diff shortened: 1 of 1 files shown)


dlemstra:
Thanks for the PR but you cannot just set the depth you will also need to store the data in that depth. And I tried that earlier but that did not seem to work with the libheif library at that point in time.

indlin:
Oh, sorry. No Simple solution? I thought, if source 16-bit, then its stored as 16 bit internally... But I don't know how Magick work inside.

dlemstra:
Closing this PR for now.

PR #4186, opened 2021-09-11
Update comment about available ImageTypes

atlesn:
Some types listed in the comment above the type set function did not match the names in the enum. I left OptimizeType and UndefinedType out from the list, hope that's correct.


dlemstra:
Thanks for helping us with this!

Issue #4319, opened 2021-10-05
libjxl error


SoftCreatR:
### ImageMagick version

7.1.0-9 Q16

### Operating system

Linux

### Operating system, version and so on

Ubuntu 20.04

### Description

When trying to process an image with the latest jpegxl version (0.6):

```
2021-10-05T22:36:26+00:00 0:00.000 0.000u 7.1.0 Configure convert[65799]: utility.c/ExpandFilenames/969/Configure
  Command line: convert {-debug} {ALL} {test.png} {test.jxl}
2021-10-05T22:36:26+00:00 0:00.001 0.000u 7.1.0 Policy convert[65799]: policy.c/IsRightsAuthorized/636/Policy
  Domain: Module; rights=Unrecognized; pattern="PNG" ...
2021-10-05T22:36:26+00:00 0:00.001 0.000u 7.1.0 Module convert[65799]: module.c/OpenModule/1273/Module
  Searching for module "PNG" using filename "png.la"
2021-10-05T22:36:26+00:00 0:00.001 0.000u 7.1.0 Module convert[65799]: module.c/GetMagickModulePath/546/Module
  Searching for coder module file "png.la" ...
2021-10-05T22:36:26+00:00 0:00.001 0.000u 7.1.0 Module convert[65799]: module.c/OpenModule/1282/Module
  Opening module at path "/usr/local/lib/ImageMagick-7.1.0/modules-Q16HDRI/coders/png.la"
2021-10-05T22:36:26+00:00 0:00.001 0.000u 7.1.0 Module convert[65799]: module.c/OpenModule/1309/Module
  Method "RegisterPNGImage" in module "PNG" at address 0x7fb7a3c10480
2021-10-05T22:36:26+00:00 0:00.001 0.000u 7.1.0 Module convert[65799]: module.c/OpenModule/1323/Module
  Method "UnregisterPNGImage" in module "PNG" at address 0x7fb7a3c10960
2021-10-05T22:36:26+00:00 0:00.002 0.000u 7.1.0 Policy convert[65799]: policy.c/IsRightsAuthorized/636/Policy
  Domain: Path; rights=Read; pattern="test.png" ...
2021-10-05T22:36:26+00:00 0:00.002 0.000u 7.1.0 Blob convert[65799]: blob.c/OpenBlob/3429/Blob
    read 3 magic header bytes
[...]

PR #4323, opened 2021-10-06
Update libjxl to use a .pc

deymo:
### Description

Starting with libjxl v0.6 we can build static or shared libs when using
libjxl, so the check for `enable_static` is not needed anymore.

Also, it is possible to just use the .pc file from libjxl.

When encoding, not all the basic_info fields were set to valid values
when using memset(). This patch uses `JxlEncoderInitBasicInfo` to
initialize them instead of memset().

### DIFF
--- a/coders/jxl.c
+++ b/coders/jxl.c
@@ -622,7 +622,7 @@ static MagickBooleanType WriteJXLImage(const ImageInfo *image_info,Image *image,
     }
   memset(&format,0,sizeof(format));
   JXLSetFormat(image,&format);
-  memset(&basic_info,0,sizeof(basic_info));
+  JxlEncoderInitBasicInfo(&basic_info);
   basic_info.xsize=(uint32_t) image->columns;
   basic_info.ysize=(uint32_t) image->rows;
   basic_info.bits_per_sample=8;
--- a/configure
+++ b/configure
@@ -816,9 +816,10 @@ LCMS_DELEGATE_FALSE
 LCMS_DELEGATE_TRUE
 LCMS2_LIBS
 LCMS2_CFLAGS
-JXL_LIBS
 JXL_DELEGATE_FALSE
 JXL_DELEGATE_TRUE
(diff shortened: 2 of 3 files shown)


coders/jxl.c
```diff
@@ -622,7 +622,7 @@ static MagickBooleanType WriteJXLImage(const ImageInfo *image_info,Image *image,
     }
   memset(&format,0,sizeof(format));
   JXLSetFormat(image,&format);
-  memset(&basic_info,0,sizeof(basic_info));
+  JxlEncoderInitBasicInfo(&basic_info);
```
dlemstra, line 625: Is this a breaking change in version `0.6`? Could this cause the issue reported in #4319?

deymo: This is sort of a breaking change, the breaking change is to reject basic_info with invalid values. The issue is that memset to 0 all fields in the basic_info is not a valid basic info because in some fields 0 is not a valid value, which in 0.6 it causes the function to return error. Before 0.6 the invalid value was silently ignored. In addition to that, while 0 is a valid value in other fields, the default is not great (example, intensity_target = 0 is not a good choice), so we now provide a function to set reasonable defaults.

dlemstra: Thanks for the info.

dlemstra:
I will need to review these changes and make sure that static linking still works. This was broken in earlier versions of jpeg-xl.

deymo:
> I will need to review these changes and make sure that static linking still works. This was broken in earlier versions of jpeg-xl.

Yes, before it was broken. The reason was because `libjxl.a` was using skcms.a which was not installed and there's no skcms package available anywhere to depend on in the `pkg-config` file. We "fixed" this by bundling the skcms.cc into libjxl.a but changing the function names to include a `jxl_` prefix to avoid potential conflict with projects also using skcms somewhere else. The real fix would be to package skcms in distributions, but that's not happening yet.

Issue #2165, opened 2020-06-17
vertical stripes in resized images


PontiacCZ:
### Description
Lanczos or Mitchell resize methods produce vertical stripes in generated images. It looks like this:

![image](https://user-images.githubusercontent.com/9153595/84920946-1f188280-b0c4-11ea-88cd-ada273d9005c.png)

### Steps to Reproduce

1. Download `tower.jpg` from [my Dropbox](https://www.dropbox.com/sh/soy1w22w8jcruba/AAAG61zUFsYdjukveBQyXbbla?dl=0)
2. run following command in the image directory: `magick tower.jpg -filter Lanczos -resize 2000x2000 tower_Lanczos.jpg`
3. examine generated image `tower_Lanczos.jpg` - contains 6 vertical stripes

### System Configuration

- ImageMagick version:
Version: ImageMagick 7.0.10-19 Q16 x64 2020-06-12 http://www.imagemagick.org
Copyright: Copyright (C) 1999-2018 ImageMagick Studio LLC
License: http://www.imagemagick.org/script/license.php
Visual C++: 180040629
Features: Cipher DPC HDRI OpenCL
Delegates (built-in): bzlib cairo flif freetype gslib heic jng jp2 jpeg lcms lqr lzma openexr pangocairo png ps raw rsvg tiff webp xml zlib

- Environment (Operating system, version and so on): Windows 10 Pro (64 bit)
- Additional information: Lenovo X395, AMD Ryzen Pro 3500U, 2.10 GHz, 16 GB RAM

fmw42:
Your command an image work fine for me on IM 7.0.10.19 Q16 Mac OSX with libjpeg 90.  What jpg delegate and version are you using? You can find that from

`magick -list format`

and look at the end of the line for JPG.

PontiacCZ:
At the _beginning_ of the line :-) . It returns following line for JPG:

`JPG* rw-   Joint Photographic Experts Group JFIF format (libjpeg-turbo 2.0.2)`

I don't understand what that is. It just all came with the portable windows package. :-) 

BTW you have Mac OSX version. Maybe the troubles are only with the Windows version.

fmw42:
My suspicion is that it is your libjpeg-turbo. You are using an older version. It is now at 2.0.4. See https://sourceforge.net/projects/libjpeg-turbo/files. Perhaps try upgrading that and see if that fixes the issue.

PontiacCZ:
[...]

PR #4421, opened 2021-10-29
Fix stripes in output when resizing with OpenCL acceleration enabled

roelandschoukens:
### Description

This patch is a possible fix for issue https://github.com/ImageMagick/ImageMagick/issues/2165.

For certain combinations of input and output size, the output contains horizontal and/or vertical stripes. These are always on the last line of a block of 32 pixels.

This patch updates the calculation of the amount of source pixels that needs to be cached to calculate the output values. This has to account for the worst case — the exact amount needed varies per block because it depends on the exact alignment of source and destination pixels (and the filter support size).

I don’t fully understand how exactly the pixel grids and filter kernels are lined up. But I have not seen any cases that break so far with this patch.

If you want to test a series of sizes on the Windows command line you can do this with a command like this:

```bat
set MAGICK_OCL_DEVICE=ON
FOR /L %a in (32, 1, 199) do magick.exe -size 200x200 gradient: -filter Lanczos -resize x%a test-%a.png
```

(n.b. in a .cmd file you have to write `%%a` instead of `%a`)

### DIFF
--- a/MagickCore/accelerate.c
+++ b/MagickCore/accelerate.c
@@ -3980,8 +3980,6 @@ static MagickBooleanType resizeHorizontalFilter(MagickCLDevice device,
     support;
 
   int
-    cacheRangeStart,
-    cacheRangeEnd,
     numCachedPixels,
     resizeFilterType,
     resizeWindowType;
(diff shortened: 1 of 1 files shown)


dlemstra (approved):

PR #4723, opened 2022-01-17
Fix wrong logic when compare two float values with '==" directly

liuchang-cc:
Analysis: {image's length call MagickWriteImages() isn't equal with call MagickGetImagesBlob(), compare two float values with == directly, it's not always expected result}
Solution: {compare two float value's difference with 1e-6 here}

### Prerequisites

### DIFF
--- a/coders/gif.c
+++ b/coders/gif.c
@@ -1729,7 +1729,7 @@ static MagickBooleanType WriteGIFImage(const ImageInfo *image_info,Image *image,
         (void) WriteBlobByte(image,(unsigned char) (opacity >= 0 ? opacity :
           0));
         (void) WriteBlobByte(image,(unsigned char) 0x00);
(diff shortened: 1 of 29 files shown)


coders/gif.c
```diff
@@ -1729,7 +1729,7 @@ static MagickBooleanType WriteGIFImage(const ImageInfo *image_info,Image *image,
         (void) WriteBlobByte(image,(unsigned char) (opacity >= 0 ? opacity :
           0));
         (void) WriteBlobByte(image,(unsigned char) 0x00);
-        if ((image->gamma != 1.0f/2.2f))
+        if (fabs(image->gamma - 1.0f/2.2f) > (1e-6))
```
dlemstra, line 1732: Thanks for the patch but can you use `MagickEpsilon` insead of `(1e-6)`?

liuchang-cc: yes, MagickEpsilon is better

Issue #4747, opened 2022-01-24
Broken build with old libheif


remicollet:
### ImageMagick version

7.1.0-20

### Operating system

Linux

### Operating system, version and so on

RHEL/CentOS 7

### Description

Using libheif 1.3.2

```
coders/heic.c:591:5: error: storage size of 'heif_filetype' isn't known
     heif_filetype;
     ^
coders/heic.c:595:3: warning: implicit declaration of function 'heif_check_filetype' [-Wimplicit-function-declaration]
   heif_filetype=heif_check_filetype(magick,(int) length);
   ^
coders/heic.c:596:24: error: 'heif_filetype_yes_supported' undeclared (first use in this function)
   if (heif_filetype == heif_filetype_yes_supported)
                        ^
coders/heic.c:596:24: note: each undeclared identifier is reported only once for each function it appears in
coders/heic.c:591:5: warning: unused variable 'heif_filetype' [-Wunused-variable]
     heif_filetype;
     ^
make[1]: *** [coders/heic_la-heic.lo] Error 1

```

### Steps to Reproduce

Build on CentOS 7 using libheif from [rpmfusion](https://rpmfusion.org/)

remicollet:
Introduced in 1b89b2b729238758a98c596c256a1d5753fa5670

remicollet:
Version 6.9.12-35 is also affected

remicollet:
API introduced in libheif 1.4.0 https://github.com/strukturag/libheif/commit/a898727c

dlemstra:
Thanks for the problem report.  We can reproduce it and will have a patch to fix it in the GIT main branch @ https://github.com/ImageMagick/ImageMagick later today.  The patch will be available in the beta releases of ImageMagick @ https://imagemagick.org/download/beta/ by sometime tomorrow.

PR #4748, opened 2022-01-24
re-add check for libheif < 1.4.0

remicollet:
see #4747


dlemstra:
See my comment on your reply. When people want to have this automatic format detection they will need to upgrade to a more recent version of libheif.

PR #5084, opened 2022-04-25
Add support for FITS images with NaN values

dgarciabriseno:
### Prerequisites

### Description
When the first pixel of a FITS image is NaN, the minima/maxima for the image ends up being set to NaN and the resulting image that's rendered is all black.

This change updates the initial minima/maxima to be DBL_MAX and DBL_MIN respectively instead of using the 1st pixel value. This allows the min/max search to work correctly, and the image is then displayed correctly.

I also added a change to override any NaN values to 0 for extra safety, since I'm not sure how the rest of ImageMagick handles NaN values.

### DIFF
--- a/coders/fits.c
+++ b/coders/fits.c
@@ -196,7 +196,7 @@ static MagickOffsetType GetFITSPixelExtrema(Image *image,
   number_pixels=(MagickSizeType) image->columns*image->rows;
   *minima=DBL_MAX;
   *maxima=DBL_MIN;
-  for (i=1; i < (MagickOffsetType) number_pixels; i++)
+  for (i=0; i < (MagickOffsetType) number_pixels; i++)
   {
     pixel=GetFITSPixel(image,bits_per_pixel);
     if (pixel < *minima)
--- a/coders/fits.c
+++ b/coders/fits.c
@@ -145,27 +145,34 @@ static MagickBooleanType IsFITS(const unsigned char *magick,const size_t length)
 
 static inline double GetFITSPixel(Image *image,int bits_per_pixel)
 {
+  double pixel = 0;
   switch (image->depth >> 3)
   {
     case 1:
-      return((double) ReadBlobByte(image));
+      pixel=((double) ReadBlobByte(image));
     case 2:
-      return((double) ((short) ReadBlobShort(image)));
+      pixel=((double) ((short) ReadBlobShort(image)));
     case 4:
     {
       if (bits_per_pixel > 0)
-        return((double) ReadBlobSignedLong(image));
-      return((double) ReadBlobFloat(image));
+        pixel=((double) ReadBlobSignedLong(image));
+      pixel=((double) ReadBlobFloat(image));
     }
     case 8:
     {
       if (bits_per_pixel > 0)
-        return((double) ((MagickOffsetType) ReadBlobLongLong(image)));
+        pixel=((double) ((MagickOffsetType) ReadBlobLongLong(image)));
     }
     default:
       break;
   }
-  return(ReadBlobDouble(image));
+  pixel=(ReadBlobDouble(image));
+
+  // Override any undefined pixels to black.
+  if (isnan(pixel))
+      pixel=0;
+
+  return pixel;
 }
 
 static MagickOffsetType GetFITSPixelExtrema(Image *image,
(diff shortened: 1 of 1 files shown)


dlemstra:
Is it possible to add an image that demonstrates this issue? 

And why do we need the changes to `GetFITSPixel`? Could we not do these checks in the `for` loop instead? And when you don't read the first pixel but set it to `DBL_MAX` and `DBL_MIN` the for loop should start at `0` instead.

dgarciabriseno:
I put the changes in GetFITSPixel since it is called later when setting pixel data on the Image object, and I wanted to include the NaN->0 there too.

You're right, the loop should start at 0, I'll patch that.

Here is a demonstration, pre-change:
image

After patch:
image

Here's the fits file I'm testing with (zipped so I can attach it here) [211_to_304.fits.zip](https://github.com/ImageMagick/ImageMagick/files/8558060/211_to_304.fits.zip)
It was generated with [sunpy](https://docs.sunpy.org/en/stable/generated/gallery/map_transformations/reprojection_different_observers.html)

coders/fits.c
```diff
@@ -145,27 +145,34 @@ static MagickBooleanType IsFITS(const unsigned char *magick,const size_t length)
...
+        pixel=((double) ((MagickOffsetType) ReadBlobLongLong(image)));
     }
     default:
       break;
   }
-  return(ReadBlobDouble(image));
+  pixel=(ReadBlobDouble(image));
+
+  // Override any undefined pixels to black.
+  if (isnan(pixel))
```
dlemstra, line 172: This should use the `IsNaN` macro instead and I am wonder if we should only do this when call to `ReadBlobDouble` is made.

coders/fits.c
```diff
@@ -145,27 +145,34 @@ static MagickBooleanType IsFITS(const unsigned char *magick,const size_t length)
 
 static inline double GetFITSPixel(Image *image,int bits_per_pixel)
 {
+  double pixel = 0;
   switch (image->depth >> 3)
   {
     case 1:
-      return((double) ReadBlobByte(image));
+      pixel=((double) ReadBlobByte(image));
```
dlemstra, line 152: You are missing `break` statements.

coders/fits.c
```diff
@@ -145,27 +145,34 @@ static MagickBooleanType IsFITS(const unsigned char *magick,const size_t length)
...
     {
       if (bits_per_pixel > 0)
-        return((double) ((MagickOffsetType) ReadBlobLongLong(image)));
+        pixel=((double) ((MagickOffsetType) ReadBlobLongLong(image)));
     }
     default:
       break;
   }
-  return(ReadBlobDouble(image));
+  pixel=(ReadBlobDouble(image));
```
dlemstra, line 169: You are now overwriting the `pixel` value that was set in the switch statement?

dgarciabriseno: 🤦🏽‍♂️ I got excited after figuring out the NaN thing and forgot how to code.

coders/fits.c
dgarciabriseno: The float could technically come back as a NaN too.
I expect any of the integer values casted to double will return false for NaN, so the override to 0 shouldn't affect them.

dgarciabriseno: Ok, the logic here got weird. I changed to only do the NaN check when using ReadBlobDouble like you suggested. That seems like the simplest solution

coders/fits.c
dgarciabriseno: Updated, returns are kept now, so this doesn't apply

coders/fits.c
dgarciabriseno: yea, changed the previous case statements to keep their returns, so pixel is only used for ReadBlobDouble here and nothing is overwritten.

dlemstra (approved):
Thanks for your contribution.

dlemstra:
While waiting on your build I decided to check if maybe we could solve this differently. The following also seems to work:

```c
  *minima=DBL_MAX;
  *maxima=DBL_MIN;
  for (i=0; i < (MagickOffsetType) number_pixels; i++)
  {
    pixel=GetFITSPixel(image,bits_per_pixel);
    if (IsNaN(pixel))
      continue;
    if (pixel < *minima)
      *minima=pixel;
    if (pixel > *maxima)
      *maxima=pixel;
  }
```

And it looks like I get the correct output image. Can you clarify why we would need to always check for a NaN inside `GetFITSPixel`? Maybe we should leave those pixels intact but don't use them to get the extrema's?

dlemstra:
ImageMagick can handle nan pixel values so I think we should only do this for the min/max checks. I don't mind making the patches after we get this merged.

dgarciabriseno:
Ok, tripped over git for a moment. Updated so the only changes are for finding the min/max pixels in the image.

@dlemstra Is this good to go?

dlemstra:
Yes it is :) Thanks for your help, will merge this after the build is completed.

PR #5119, opened 2022-05-05
autotools: Add ws2_32 library with MagickCore for Win32 platform

Biswa96:
### Prerequisites

### Description

Otherwise, the build process fails with the following errors in mingw toolchain:
```
ld.exe: MagickCore/.libs/libMagickCore_7_Q16HDRI_la-nt-base.o:nt-base.c:(.text+0x1d6d): undefined reference to '__imp_WSAStartup'
ld.exe: MagickCore/.libs/libMagickCore_7_Q16HDRI_la-nt-base.o:nt-base.c:(.text+0x3554): undefined reference to '__imp_WSACleanup'
```


dlemstra:
Would it be possible to add something to the build so we can make sure this stays fixed?

Biswa96:
My apology. I can not understand your question. Can you suggest any equivalent code change?

dlemstra:
I am wondering if could we run `ld.exe` during a GitHub actions build to make sure this stays fixed?

Biswa96:
If I understand correctly, you want to verify this in CI because it did not catches the error before, right? Let me be familiar with how CI works here.

dlemstra:
Yes that is the point that I am trying to make. And I would prefer to have another PR that adds this first before merging this one.

dlemstra:
Thanks for that other PR to remove the workaround. Would it also possible to create a PR where the build breaks because ld.exe fails?

Biswa96:
Wouldn't it be same if you restart msys2 CI from any commit in this repository? Assuming that wouldn't make things worse. Then restart the msys2 CI here.

dlemstra:
Yeah that actually seems to happen so we are good to go to get this merged if the build passes again in this PR.

dlemstra:
It looks like this still fails with your patches.

Biswa96:
Just realized, does the CI regenerate Makefile.in, configure etc. files? This change may need to regenerate the files. I can not do that in msys2 because the autotools version mismatch with current one.

dlemstra:
How did you test this locally? And why can we not run the autotools in GitHub actions?

Biswa96:
> How did you test this locally?

Biswa96:
> How did you test this locally?

I regenerated those autotools files in msys2. But the commands will bump the required autotools versions. Some prefer to stick with older autotools version for backward compatibility. So, I did not add those changes in this commit.

> And why can we not run the autotools in GitHub actions?

It is possible to do so but falls under same reason as above.

dlemstra:
It should not be an issue that the version is bumped because we don't commit any files during the build.

Biswa96:
OK, let me test in my fork first.

dlemstra:
It looks like your patch fixes the build. Thanks for helping us out with this.

Issue #2788, opened 2020-11-03
Reading animated AVIF fails


andrius4669:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am using the latest version of ImageMagick
- [x] I have searched [open](https://github.com/ImageMagick/ImageMagick/issues) and [closed](https://github.com/ImageMagick/ImageMagick/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Trying `magick identify` on files in https://github.com/AOMediaCodec/av1-avif/tree/master/testFiles/Netflix/avis fails with errors like
```
identify: Invalid input: Non-existing item ID referenced: Non-existing alpha image referenced `alpha_video.avif' @ error/heic.c/IsHeifSuccess/134.
```
or
```
identify: Unsupported file-type: Unspecified: File does not include any supported brands.
 `Chimera-AV1-10bit-480x270.avif' @ error/heic.c/IsHeifSuccess/134.
```
Both of these open fine in gwenview (with qt-avif-image-plugin installed) and also chromium.

### Steps to Reproduce

```
git clone https://github.com/AOMediaCodec/av1-avif
cd av1-avif/testFiles/Netflix/avis
magick identify alpha_video.avif
magick identify Chimera-AV1-10bit-480x270.avif
```

### System Configuration

- ImageMagick version: 7.0.10-35 Q16 x86_64 2020-11-01
- Environment (Operating system, version and so on): x86_64 ArchLinux
- Additional information: libheif 1.9.1, libavif 0.8.2, dav1d 0.7.1, aom 2.0.0

EDIT: adjusted for https://github.com/AOMediaCodec/av1-avif/commit/a33fd7ba8d4d19bef2a59d843cc61f0642296d92 and omitted details only relevant for `.avifs` file extension workaround since upstream now uses just `.avif`
Archive of relevant files for record purposes: [imgs.zip](https://github.com/ImageMagick/ImageMagick/files/5540075/imgs.zip) https://github.com/AOMediaCodec/av1-avif/tree/a33fd7ba8d4d19bef2a59d843cc61f0642296d92/testFiles/Netflix/avis

urban-warrior:
Can you verify your link?  We get a 404 on the link you provided.

andrius4669:
[...]

PR #5327, opened 2022-07-14
resolve #5241

unclesamwk:
### Prerequisites

- [ ] I have written a descriptive pull-request which add avif(rw)/heic(r) support

### DIFF
--- a/.github/workflows/app-image.yml
+++ b/.github/workflows/app-image.yml
@@ -30,7 +30,7 @@ jobs:
       run: |
         set -e
         apt-get update -y
-        apt-get install -y autoconf curl fuse git kmod libbz2-dev libdjvulibre-dev libfontconfig-dev libfreetype6-dev libfribidi-dev libharfbuzz-dev liblcms-dev libopenexr-dev libturbojpeg0-dev liblqr-dev libraqm-dev libtiff-dev libwebp-dev libx11-dev libxml2-dev liblzma-dev software-properties-common wget
+        apt-get install -y autoconf curl fuse git kmod libbz2-dev libdjvulibre-dev libfontconfig-dev libfreetype6-dev libfribidi-dev libharfbuzz-dev liblcms-dev libopenexr-dev libturbojpeg0-dev liblqr-dev libraqm-dev libtiff-dev libwebp-dev libx11-dev libxml2-dev liblzma-dev software-properties-common wget libtool
         add-apt-repository ppa:git-core/ppa -y
         apt-get install -y git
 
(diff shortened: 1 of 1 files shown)


dlemstra:
Why are you building this from source?

dlemstra:
I think we should install packages instead of building them. Maybe we could change the container image to a more recent version of Ubuntu to get a more recent version?

unclesamwk:
When i use ubuntu 20.04 or 22.04 linuxdeployqt doesn't work.

```
| linuxdeployqt  (commit d6ac06c), build 19 built on 2022-07-04 19:17:26 UTC
| ERROR: The host system is too new.
| Please run on a system with a glibc version no newer than what comes with the oldest
| currently still-supported mainstream distribution (Ubuntu Bionic), which is glibc 2.27.
| This is so that the resulting bundle will work on most still-supported Linux distributions.
| For more information, please see
| https://github.com/probonopd/linuxdeployqt/issues/340
```
https://github.com/probonopd/linuxdeployqt/issues/340
https://github.com/probonopd/linuxdeployqt/issues/536

unclesamwk:
I have now split the compile and packing appimage into 2 different jobs. Unfortunately, I can not test this locally with act and the articats reasonably. Anyone else?

dlemstra:
I did some reading this morning and the advice is to build on an older version of Ubuntu to make it work on older versions of Ubuntu. I am wondering if we could do what is mentioned in "Recommended alternative to pinning" described here: https://help.ubuntu.com/community/PinningHowto. This might make it possible to use a more recent version of libheif. And if that doesn't work I think we should install and use the older version instead for now. And "upgrade" to 20.04 once app-image uses a more recent version.

unclesamwk:
I have tried pinning, but this requires all dependencies to be resolved and handled differently by pinning. I think that is unnecessary more effort. I personally don't understand where the problem is now. If an appimage file is successfully built, and everything works as expected, then you would have an upgrade from IM to 22.04 ready to go. Whether the build of the appimage is now done with 18.04 or later in higher versions has, does little to the matter.

Sondro:
> I have tried pinning, but this requires all dependencies to be resolved and handled differently by pinning. I think that is unnecessary more effort. 

Unsure if you noticed, but under the page [dlemstra](https://github.com/dlemstra) referenced under 'Recommended **alternative to pinning**
' it mentions an alternative because pinning breaks & fails using `apt-get build-dep package` & `apt-get -b source -t package`.

Anyways, if this is going to take a while it would be nice to have animated AVIF support -- which would also cover #2788: https://github.com/ImageMagick/ImageMagick/issues/2788#issuecomment-727815710

unclesamwk:
In ubuntu 18.04 the "initrip" doesn't really work either:
```
Err:6 http://archive.ubuntu.com/ubuntu intrepid Release
  404 Not Found [IP: 185.125.190.36 80]
```
It doesn't make sense to install on a not up to date ubuntu release.

btimbermont:
@Sondro Could you please check the other PR that adds HEIC support to the app image?
It keeps using the old Ubuntu version as recommended for bnuilding app images, and  I have tested the github workflow and it produces a fully functional appimage with HEIC support.

The PR is this one:  https://github.com/ImageMagick/ImageMagick/pull/6098

dlemstra:
Closing this now because #6098 added support for this.

PR #5452, opened 2022-08-17
Support DDS file DDPF_LUMINANCEA type of data

LoserFox:
When the surface data type of DDS texture is DDPF_LUMINANCE and DDPF_ALPHAPIXELS combination dual-channel file, "ImageTypeNotSupported" will appear on the console, and I've modified some of the code so that ImageMagick can handle DDS textures of that type

### DIFF
--- a/coders/dds.c
+++ b/coders/dds.c
@@ -2390,6 +2390,11 @@ static MagickBooleanType ReadUncompressedRGBAPixels(Image *image,
           alphaBits=2;
           (void) SetImageType(image,GrayscaleAlphaType,exception);
         }
+        else if (IsBitMask(dds_info->pixelformat,0x00ff,0x0000,0x0000,0xff00))
+        {
+          alphaBits=2;
+          (void) SetImageType(image,GrayscaleAlphaType,exception);
+        }
       else if (IsBitMask(dds_info->pixelformat,0x0f00,0x00f0,0x000f,0xf000))
         alphaBits=4;
       else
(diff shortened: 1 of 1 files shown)


dlemstra:
Can you please explain why this is the proper fix with a reference to some kind of document/spec that explains this?

LoserFox:
This is a a document provided by the Microsoft's official, it developed some rules of the DDS texture, https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dds-pixelformat
In addition, I provide a DDPF_LUMINANCEA (DDPF_LUMINANCE+DDPF_ALPHAPIXELS) texture file as a sample
[texture.zip](https://github.com/ImageMagick/ImageMagick/files/9362031/texture.zip)

coders/dds.c
```diff
@@ -2390,6 +2390,11 @@ static MagickBooleanType ReadUncompressedRGBAPixels(Image *image,
           alphaBits=2;
           (void) SetImageType(image,GrayscaleAlphaType,exception);
         }
+        else if (IsBitMask(dds_info->pixelformat,0x00ff,0x0000,0x0000,0xff00))
```
dlemstra, line 2393: Could you combine this with an `||` in the previous `else if`?

dlemstra (approved):
Thanks for your contribution and explanation.

dlemstra:
Going to apply the patch myself. Thanks for your contribution.

PR #5538, opened 2022-09-07
Fix HEIF header include path for MINGW

kmilos:
### Prerequisites

### Description
MINGW packages install headers into `${includedir}/libheif` just like UNIX.

In fact, [even w/ MSVC they should go into a subdirectory like this](https://github.com/strukturag/libheif/blob/1d36c10ade9b2acfc2dcc4ca8b6176a9789c902c/libheif/CMakeLists.txt#L165) (haven't checked e.g. vcpkg to confirm), so not sure why this check is this still needed?


dlemstra:

PR #5713, opened 2022-11-03
utilities: Enable support for unicode paths in mingw

Biswa96:
### Prerequisites

### Description

This fixes magick.exe with unicode file name. Otherwise the command fails as following:

```
$ magick identify páramo.png
identify: unable to open image 'páramo.png': No such file or directory @ error/blob.c/OpenBlob/3570.
```

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -259,6 +259,7 @@ check_PROGRAMS = $(am__EXEEXT_2) $(am__EXEEXT_4)
 @WIN32_NATIVE_BUILD_TRUE@am__append_1 = -lws2_32
 @MAGICKCORE_ZERO_CONFIGURATION_SUPPORT_TRUE@am__append_2 = MagickCore/threshold-map.h
 @MAGICKCORE_ZERO_CONFIGURATION_SUPPORT_TRUE@am__append_3 = MagickCore/threshold-map.h
+@WIN32_NATIVE_BUILD_TRUE@@WITH_UTILITIES_TRUE@am__append_4 = -municode
 subdir = .
 ACLOCAL_M4 = $(top_srcdir)/aclocal.m4
 am__aclocal_m4_deps = $(top_srcdir)/m4/ac_func_fseeko.m4 \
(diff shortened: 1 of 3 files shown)


Biswa96:
@urban-warrior Some portion of this commit was changed in fb70c5b55254484b9b2a2e0b5393a9e90b6108f6. That commit broke mingw build. Please revert that change, if possible.

urban-warrior:
Attempting to move the check for unicode into the configure script rather than in the Makefile.  We'll make another attempt, and if it fails, we'll revert to using your LDFLAGS_EXTRA codeset.

sergeevabc:
Err… Fellows, years are passing by as well as issues about Unicode, but
on Windows we still have the following anomaly:
```console
$ magick convert -monitor C:\Квартира\комната.jpg out.pdf
load image[C:\╨Ъ╨▓╨░╤А╤В╨╕╤А╨░\╨║╨╛╨╝╨╜╨░╤В╨░.jpg]: 1279 of 1280, 100% complete
...
```
Any chance to fix that?

fmw42:
See https://imagemagick.org/Usage/text/#unicode

sergeevabc:
@fmw42, not sure what should I grasp from that very long page. I have lots of modern command-line tools (ffmpeg, ripgrep, curl to name a few) which output non-Latin symbols correctly, without giving them a gibberish look. So it's possible. Question is when IM will join the club?

fmw42:
Type the symbols using a UTF-8 compatible text editor or type them into a UTF-8 compatible text file and then put the text file into the place for the text as @textfile.txt. See example shown on that page. The important thing is that the editor or file must be UTF-compatible

urban-warrior:
The progress monitor directly takes the image filename and prints it.  That works fine for UTF-8 but if its a wide-character stream, it does not print properly under Windows.  ImageMagick honors UTF-8 encoding by default.  It tries to makes accommodations, where possible, for wide-characters.  We'll investigate and try to commit a patch once we find a proper solution to the issue.

sergeevabc:
@fmw42, I'm talking not about producing an image with Unicode symbols, which is obvious from code examples I provided, but about the distorted look of paths and filenames while ImageMagick process them. To reproduce the issue, just create folder with Cyrillic (Russian) letters `C:\Квартира`, place some image there named `комната.jpg`, then run the usual shell `cmd.exe` and launch `magick.exe convert -monitor C:\Квартира\комната.jpg out.pdf` only to see that processing goes fine except that path and filenames turns into `╨Ъ╨▓╨░╤А╤В╨╕╤А╨░\╨║╨╛╨╝╨╜╨░╤В╨░.jpg`. It means that ImageMagick is still not fully Unicode compatible internally, hence the issue.

yarusome:
Currently, invoking `magick` in `cmd.exe` automatically changes the code page to 65001 (at least on Windows 8.1), but this is causing other problems in the console *if* the default code page is not 65001, such as making contents displayed normally before invoking `magick` gibberish.

Other utils, such as FFmpeg (like what sergeevabc pointed out), Aria2, Ghostscript, to name a few, don't exhibit such issues, which implies that there should be a generic solution.

Issue #4375, opened 2021-10-15
sRGB images are written to PNG without an sRGB chunk, and `identify -verbose` does not make that apparent


sboukortt:
### ImageMagick version

7.0.10-11

### Operating system

Windows

### Operating system, version and so on

Windows 10 20H2, MINGW64 (MSYS2), ImageMagick 7.0.10-11 Q16 x86_64 2021-07-24

### Description

For sRGB images, ImageMagick does not seem to include an sRGB chunk in the PNG files that it creates. It does include the [“fallback” cHRM and gAMA chunks](http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#:~:text=An%20application%20that%20writes%20the%20sRGB%20chunk%20should%20also%20write%20a%20gAMA%20chunk%20(and%20perhaps%20a%20cHRM%20chunk)%20for%20compatibility%20with%20applications%20that%20do%20not%20use%20the%20sRGB%20chunk.), which however makes such images look different since the sRGB transfer function is not exactly a gamma curve and gAMA=45455 is only an approximation.

Attached are four images: one with no color space chunk at all, one with just an sRGB chunk, one with sRGB + gAMA/cHRM (which should look identical to the one with just the sRGB chunk since sRGB takes precedence), and one with just gAMA/cHRM (which should look slightly different). Google Chrome indeed displays the first three identically, and the fourth one with slightly darker shadows. Their chunks can be verified with `pngcheck -v` or with [TweakPNG](http://entropymine.com/jason/tweakpng/).

But when running `convert input.png output.png` on them, they all end up with gAMA/cHRM and no sRGB chunk. A fact that `identify -verbose` does not make clear at all since it nevertheless says:

```
png:sRGB: intent=0 (Perceptual Intent)
```

### Steps to Reproduce

The `identify -verbose` issue:

`identify -verbose images/image-sRGB.png` and `identify -verbose images/image-gAMA-cHRM.png` display apparently identical pixel value statistics and image properties, and yet those two images look different (as they should) when opened in Chrome. `pngcheck -v` displays their actual chunks.

----

The `convert` issue:

After running `mkdir -p output; parallel convert '{}' output/'{/}' ::: images/*.png`, the output images all end up with a gAMA and cHRM chunks but no sRGB chunk. If the input images had one, it is lost, and thus the image is displayed differently by color-managed viewers that take the gAMA chunk into account.

[...]

PR #5824, opened 2022-11-28
Chrome, Mozilla, GIMP do not assume sRGB

ValZapod:
I think it was misworded, 2.2 gAMA should be deleted, cause it is mandatery in the case sRGB chunk is included, but does not mean sRGB if sRGB chunk is absent. It means pure 2.2 gamma. Doc part of #4375.

Reference: PNG spec. about fallback for sRGB chunk.

Photoshop still cannot read any chunks.

GIMP added proper support for chunks per my many questions.

### Prerequisites

### DIFF
--- a/coders/png.c
+++ b/coders/png.c
@@ -2920,26 +2920,6 @@ static Image *ReadOnePNGImage(MngInfo *mng_info,
           ping_found_sRGB_cHRM=MagickTrue;
     }
 
-  if (image->rendering_intent != UndefinedIntent)
-    {
-      if (ping_found_sRGB != MagickTrue &&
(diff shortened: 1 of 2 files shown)


ValZapod:
Second commit removes logic that PNG is assumed when cHRM and gAMA fallbacks are used.

ValZapod:
To verify that Gamma 2.2, used for BT.709 gamma 2.4 and sRGB are NOT the same see Davinci resolve or this: http://cinelex.ru/blog/sravnenie-gammy-v-raznyx-standartax/ (or see [the open bug](https://phabricator.wikimedia.org/T26768) in Wikipedia engine)


image
image

PR #5929, opened 2023-01-03
Fix Exif writing for HEIF/JPEG XL

kmilos:
### Prerequisites

### Description
The HEIF method actually takes care of the offset calculation, and there is no limit on Exif size for HEIFs.


kmilos:
@urban-warrior This is still not working for JPEG XL, I guess `StringToStringInfo("\0\0\0\6")` is not considered a string of length 4 but an empty string?

PR #6093, opened 2023-02-21
Remove fix for BeOS r4 which breaks gcc build with external toolchain

eitlane:
Hi all, 

This piece of code seems obsolete as it fixed something in BeOS R4 (which is from 1998). However it breaks the tests if the project is build with exteranl toolchain (not the one that comes from the Linux distribution). As LD_LIBRARY_PATH is already set (couple of lines above), $SED (sed) loads different glibc causing it to core. This leaves LD_LIBRARY_PATH *NOT* properly set and once the test are run (with make check) libmagic is unable to find external dependencies like bzip2, zlib etc. 

As I saw that $SED is used after as well it may potentially core at some other places (which however didn't happen in my build as it seems it didn't pass from that part of the code). This wouldn't happen of course if sed is completely statically build.

Thanks

### Prerequisites

### DIFF
--- a/config/ltmain.sh
+++ b/config/ltmain.sh
@@ -5731,10 +5731,6 @@ func_exec_program ()
     # Add our own library path to $shlibpath_var
     $shlibpath_var=\"$temp_rpath\$$shlibpath_var\"
 
-    # Some systems cannot cope with colon-terminated $shlibpath_var
-    # The second colon is a workaround for a bug in BeOS R4 sed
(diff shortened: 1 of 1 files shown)


urban-warrior:
This patch needs to be pushed upstream to the GNU Libtool maintainers.  Its a generated file per the header:

```
#! /usr/bin/env sh
## DO NOT EDIT - This file generated from ./build-aux/ltmain.in
##               by inline-source v2019-02-19.15

# libtool (GNU libtool) 2.4.7
# Provide generalized library-building support services.
```

eitlane:
I see. Thanks for your update. I will close the PR

Issue #4666, opened 2022-01-05
Appimage with HEIC support (and other delegates)


richard-llmnn:
### Is your feature request related to a problem? Please describe.

On the [official download page](https://imagemagick.org/script/download.php) of ImageMagick there is only one  appimage version available. The appimage supports the following delegates: `bzlib djvu fontconfig freetype jbig jng jpeg lcms lqr lzma openexr png raqm tiff webp x xml zlib`.

I have a system where no package manager or similar is available. But I need HEIC support.

### Describe the solution you'd like

Is it possible to publish two appimages?
The current one as "lite/simple version" and a second one "full version" that support  almost all image formats/delegates?
Delegates for the "full version" which are missing in the "lite/simple version":
- Autotrace          --with-autotrace
- DPS                  --with-dps
- FFTW               --with-fftw
- FLIF                  --with-flif
- FlashPIX           --with-fpx
- Ghostscript lib   --with-gslib
- Graphviz           --with-gvc
- HEIC                 --with-heic
- LTDL                 --with-ltdl
- Magick++          --with-magick-plus-plus
- OpenJP2          --with-openjp2
- PANGO            --with-pango
- PERL                --with-perl
- RAW                 --with-raw
- RSVG               --with-rsvg
- WMF                 --with-wmf

### Describe alternatives you've considered

In my opinion it is the easiest way to create two appimages.

dlemstra:
In case anyone wants to pick this up this can be done in the `app-image.yml` file.

btimbermont:
I would love this to be added

btimbermont:
@dlemstra I added HEIC support to the AppImage in this PR: https://github.com/ImageMagick/ImageMagick/pull/6098

The PR is waiting for approval/feedback though. I'd love for someone to have a look, it's a very small change.

btimbermont:
With PR  #6098 support for HEIC (and AVIF) is now added

richard-llmnn:
Thank you @btimbermont and @dlemstra for your help :100: .
Because not all formats are added yet to the current AppImage, I will keep the issue open.

btimbermont:
I added jp2 support in https://github.com/ImageMagick/ImageMagick/pull/6630

PR #6098, opened 2023-02-22
Add HEIC support to AppImage (Partially solve #4666)

btimbermont:
### Prerequisites

- [x] I have verified that there are no overlapping [pull-requests](https://github.com/ImageMagick/ImageMagick/pulls) open --> **there is one, but it doesn't work right now, has conflicts with main, has been stale for over 6 months, and has issues this PR does not**

### Description
The AppImage release of ImageMagick doesn't support the HEIC image format because the required dependency (libheif-dev) was not installed.
This PR adds libheif-dev as a dependency in the AppImage workflow, but to do so it also adds two new PPAs, both from the struktur AG team. Without these PPAs, Ubuntu 18 only offers libheif version 1.1, while ImageMagick needs v 1.4 at the least.
Alternatively libheif could be compiled from source, but this would add much buildtime.
Alternatively we could also update the base image while building to Ubuntu 20  (or 22), but this is incomaptible with the AppImage build step, which requires the Ubuntu version to be the oldest currently supported LTS release (which is 18 atm).

(Please ignore the commits in which I was messing around, I'm not that familiar with the code so I poked around a lot to see what does and doesn't work ;) I also tried adding AVIF support but that didn't work so perhaps that's for another PR)


dlemstra:
Would it be possible to do another merge of the master so we can trigger the app-image build and test it before merging?

btimbermont:
> Would it be possible to do another merge of the master so we can trigger the app-image build and test it before merging?

Done!

dlemstra (approved):
Thanks for your help and it looks like this works according to the output of the build.

PR #6230, opened 2023-04-05
PixelChannel: Add field for invalid Channel

LinusDierheimer:
### Prerequisites

### Description

The current code in `pixel-accessor.h` gives a compile warnings:
```
/usr/include/ImageMagick-7/MagickCore/pixel-accessor.h:135:26: warning: comparison is always false due to limited range of data type [-Wtype-limits]
  135 |   if (((ssize_t) channel < 0) || ((ssize_t) channel >= MaxPixelChannels))
```
making the header impossible to include, when using `-Werror`.

The warning is thrown, because the enum `PixelChannel` only has unsigned values. I therefore removed the unnecessary checks.

### DIFF
--- a/MagickCore/pixel.h
+++ b/MagickCore/pixel.h
@@ -94,7 +94,8 @@ typedef enum
   MetaPixelChannels = 10,
   IntensityPixelChannel = MaxPixelChannels,  /* ???? */
   CompositePixelChannel = MaxPixelChannels,  /* ???? */
-  SyncPixelChannel = MaxPixelChannels+1      /* not a real channel */
+  SyncPixelChannel = MaxPixelChannels+1,     /* not a real channel */
+  InvalidChannel = -1
 } PixelChannel;  /* must correspond to ChannelType */
 
 typedef enum


dlemstra:
I am not sure which compiler is giving you that warning but this is allowed in Visual Studio 2022 where I have enabled warnings are treated as errors and it actually passes `-1` to the function:

```C
GetPixelChannelTraits(image, (PixelChannel) -1);
```

LinusDierheimer:
> I am not sure which compiler is giving you that warning
```Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/12.2.1/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++,d --enable-bootstrap --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --with-build-config=bootstrap-lto --with-linker-hash-style=gnu --with-system-zlib --enable-__cxa_atexit --enable-cet=auto --enable-checking=release --enable-clocale=gnu --enable-default-pie --enable-default-ssp --enable-gnu-indirect-function --enable-gnu-unique-object --enable-libstdcxx-backtrace --enable-link-serialization=1 --enable-linker-build-id --enable-lto --enable-multilib --enable-plugin --enable-shared --enable-threads=posix --disable-libssp --disable-libstdcxx-pch --disable-werror
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 12.2.1 20230201 (GCC) 
```
After some investigation, i also have `-Wextra` passed to it, which enables `-Wtype-limits`.

> it actually passes -1 to the function

Then this patch is defective, sorry for that. I will look into disabling the warning in another way.

LinusDierheimer:
The standard states:
> A value of integral or enumeration type can be explicitly converted to a complete enumeration type. [...] If the enumeration type does not have a fixed underlying type, the value is unchanged if the original value is within the range of the enumeration values (9.7.1), and otherwise, the behavior is undefined.

So passing `-1` is undefined behaviour. Maybe that is why GCC is ignoring it and says that all *valid* values are >= 0.

urban-warrior:
Any new enums must appear at the end of the list to ensure we don't break the ABI.

PR #6940, opened 2023-12-12
support building plugins without ltdl

hiirotsuki:
### Prerequisites

### Description

support building imagemagick with shared modules without using ltdl, allows building imagemagick with slibtool and for platforms that ltdl does not support

missing configure changes

### DIFF
--- a/MagickCore/module.c
+++ b/MagickCore/module.c
@@ -67,7 +67,10 @@
 #include "MagickCore/utility.h"
 #include "MagickCore/utility-private.h"
 #if defined(MAGICKCORE_MODULES_SUPPORT)
-#if defined(MAGICKCORE_LTDL_DELEGATE)
+#if defined(MAGICKCORE_DLOPEN)
+#include <dlfcn.h>
+typedef void *ModuleHandle;
+#elif defined(MAGICKCORE_LTDL_DELEGATE)
 #include "ltdl.h"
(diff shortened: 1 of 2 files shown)


dlemstra:
Not sure if we should support this? Do you have examples of platforms where this now doesn't work? And if we do this it might be better to create new defines that either use `dlclose` or `lt_dlclose` and do the same for the other methods like `dlerror`?

Redfoxymoon:
[Midipix](https://midipix.org/), a POSIX layer for microsoft windows lacks support for ltdl, we already have imagemagick running there with dynamic plugins.

a set of macros would remove the need for most of the ifdef preprocessor changes, true; that's another option

dlemstra:
This would also require changes to the `configure.ac` file to detect either `libltdl` or `dlfcn` and sets flags accordingly. And I think `libltdl` should be preferred when that could be found.

Redfoxymoon:
Defaulting to `libltdl` would be fine

orbea:
@dlemstra Would this be acceptable with a set of --enable/--disable configure flags as suggested by @Redfoxymoon?

This would be helpful on Gentoo when using slibtool instead of GNU libtool which doesn't install the libtool archive (.la) files by default.

mjclo:
Elena 20180510_122511 (7)

PR #6942, opened 2023-12-12
opencl.c: replace lt_dlclose() with dlclose() / FreeLibrary()

valgur:
### Prerequisites

### Description
`opencl.c` makes use of the OpenCL runtime library and opens it using `dlopen()` / `LoadLibraryA()`:
https://github.com/ImageMagick/ImageMagick/blob/54b13e91d262b1083e27fc8c02532c89d3ff649c/MagickCore/opencl.c#L2495-L2499
but closes it using `lt_dlclose()`:
https://github.com/ImageMagick/ImageMagick/blob/54b13e91d262b1083e27fc8c02532c89d3ff649c/MagickCore/opencl.c#L2601-L2603

I wonder if there's a particular reason the libtool version of the `dlclose()` function is being used? Eliminating the libtool dependency simplifies Windows builds of ImageMagick quite a bit.

### DIFF
--- a/MagickCore/opencl.c
+++ b/MagickCore/opencl.c
@@ -90,9 +90,6 @@
 #include "MagickCore/utility-private.h"
 
 #if defined(MAGICKCORE_OPENCL_SUPPORT)
-#if defined(MAGICKCORE_LTDL_DELEGATE)
-#include "ltdl.h"
-#endif
 
 #ifndef MAGICKCORE_WINDOWS_SUPPORT
 #include <dlfcn.h>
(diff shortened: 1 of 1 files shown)


dlemstra:
Thanks for point this out but your solution is not the one that I would prefer to use. I just made this commit: https://github.com/ImageMagick/ImageMagick/commit/da099f2d2f62613644a942a8369bf986d9bfdf09 to make sure that this part is in line with the rest of the project. This also means that I will be closing this PR.

PR #6978, opened 2023-12-27
configure.ac: fixup LFS check for autoconf-2.72

thesamesam:
### Prerequisites

### Description

There are two problems with the check:
1) The 'unknown' case has a problem which is rejected by stricter C compilers because it has -Wimplicit-int and -Wimplicit-function-declaration warnings. Fix that.

2) For the 'other' case, we were using the value of ac_cv_sys_file_offset_bits for -D_FILE_OFFSET_BITS to pass down into the Perl module build, but autoconf-2.72 drops the use of ac_cv_sys_file_offset_bits in cf09f48841b66fe76f606dd6018bb3a93242a7c9, so this ends up defining '-D_FILE_OFFSET_BITS=' which then breaks the build.

I've added a hack for 2) to preserve the old behavior.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -503,7 +503,8 @@ if test "$enable_largefile" != no; then
     unknown)
         AC_MSG_CHECKING([for native large file support])
         AC_RUN_IFELSE([AC_LANG_PROGRAM([#include <unistd.h>
-          main () {
+          #include <stdlib.h>
+          int main () {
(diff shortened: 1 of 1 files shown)


thesamesam:
Sorry I didn't notice this before, as it only shows up with the Perl module build.

thesamesam:
Thank you!

PR #7008, opened 2024-01-06
Make Magick++-config work from non-standard places

### LINKED ISSUE #7007: Magick++-config fails when the .pc file is not installed in a place searched by pkg-config by default
### ImageMagick version

7.1.1-25

### Operating system

Linux

Fedora 39

### Description

Magick++-config fails when the .pc file is not installed in a place searched by pkg-config by default.

### Steps to Reproduce

```
$ ./configure CFLAGS=-O3 CXXFLAGS=-O3 --disable-static --prefix ~/local
$ make
$ make install
$ Magick++-config --libs
Package Magick++-7.Q16HDRI was not found in the pkg-config search path.
Perhaps you should add the directory containing `Magick++-7.Q16HDRI.pc'
to the PKG_CONFIG_PATH environment variable
Package 'Magick++-7.Q16HDRI', required by 'virtual:world', not found
```

### Images

_No response_

TedLyngmo:
When the .pc file is installed in a location not searched by pkg-config by default, Magick++-config failed.

Fixes #7007

### DIFF
--- a/Magick++/bin/Magick++-config.in
+++ b/Magick++/bin/Magick++-config.in
@@ -7,6 +7,7 @@
 
 prefix=@prefix@
 exec_prefix=@exec_prefix@
+export PKG_CONFIG_LIBDIR="${exec_prefix}/lib/pkgconfig"
 
 usage='Usage: Magick++-config [--cppflags] [--cxxflags] [--exec-prefix] [--ldflags] [--libs] [--prefix] [--version]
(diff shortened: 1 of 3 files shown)


remicollet:
This breaks common usage on every distro that didn't use "lib" directory (ex: /usr/lib64 on RPM disto)

See PR #7013

PR #7013, opened 2024-01-08
fix PKG_CONFIG_LIBDIR

remicollet:
broken by PR #7008 for distro not using `lib`, but `lib64`

### DIFF
--- a/Magick++/bin/Magick++-config.in
+++ b/Magick++/bin/Magick++-config.in
@@ -7,7 +7,9 @@
 
 prefix=@prefix@
 exec_prefix=@exec_prefix@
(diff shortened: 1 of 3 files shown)


remicollet:
Result is 
```
prefix=/usr
exec_prefix=/usr
libdir=/usr/lib64
pkgconfigdir=${libdir}/pkgconfig
export PKG_CONFIG_LIBDIR="${pkgconfigdir}"
```

TedLyngmo:
My bad! I've verified that this patch works in my non-standard location too:
```
prefix=/home/ted/local
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
pkgconfigdir=${libdir}/pkgconfig
export PKG_CONFIG_LIBDIR="${pkgconfigdir}"
```

PR #7068, opened 2024-01-25
Fix leak when throwing in `Magick::Color` ctor

### LINKED ISSUE #7066: leaking memory when `Magick::Color` throws
### ImageMagick version

7.1.1-26

### Operating system

Linux

Ubuntu 22.04

### Description

The following code leaks memory while leaving `c` equal to `nullptr`

```cpp
#include <Magick++.h>

int main() {
	Magick::Color *c = nullptr;
	try {
		c = new Magick::Color{"invalid"};
	} catch (...) {}
	printf("c = %p\n", c);
}
```

### Steps to Reproduce

Build the above program with a leak detector

### Images

_No response_

mmomtchev:
### Prerequisites

### Description

Fix #7066 

This the minimal fix. It is not clear if an assignment should invalidate the target when throwing, but here this is already the case anyway.

The right solution would be to complete eliminate all forms of a constructor initializing a member pointer with `new` (there are many cases in `Color`, and a few more elsewhere) - either by replacing `_pixel` with a member variable or with a smart pointer (this is the official current C++ guidelines: https://isocpp.org/wiki/faq/exceptions#selfcleaning-members ). As both require very extensive changes, I prefer to submit a very small patch.

### DIFF
--- a/Magick++/lib/Color.cpp
+++ b/Magick++/lib/Color.cpp
@@ -239,7 +239,12 @@ const Magick::Color& Magick::Color::operator=(const std::string &color_)
       setPixelType(target_color);
     }
   else
-    _isValid = false;
+    {
+      _isValid = false;
+      _pixelOwn = false;
+      delete _pixel;
+      _pixel = nullptr;
+    }
   ThrowPPException(false);
 
   return(*this);


dlemstra:
Thanks for helping us with this and creating this PR so quickly.

PR #7095, opened 2024-02-08
Add heic:chroma-downsampling-algorithm define for HEIC format images.

njdoyle:
### Prerequisites

### Description
This change exposes the options made available by libheif for controlling how chroma subsampling is performed. A new define is added named "heic:chroma-downsampling-algorithm" which has valid values of "nearest-neighbor", "average", and "smart-yuv". "smart-yuv" is only available if libheif was built with the "WITH_LIBSHARPYUV" compile-time option.

The "heic:chroma" define has been made available for all HEIC images instead of just AVIF images.

Version checks for libheif have been clarified and also made more accurate.

### DIFF
--- a/coders/heic.c
+++ b/coders/heic.c
@@ -347,6 +347,11 @@ static MagickBooleanType ReadHEICImageHandle(const ImageInfo *image_info,
   struct heif_image
     *heif_image;
 
+#if LIBHEIF_NUMERIC_VERSION >= HEIC_COMPUTE_NUMERIC_VERSION(1,16,0)
+  const char
+    *option;
+#endif
+
   /*
     Read HEIC image from container.
   */
@@ -376,6 +381,28 @@ static MagickBooleanType ReadHEICImageHandle(const ImageInfo *image_info,
   if (status == MagickFalse)
     return(MagickFalse);
   decode_options=heif_decoding_options_alloc();
+#if LIBHEIF_NUMERIC_VERSION >= HEIC_COMPUTE_NUMERIC_VERSION(1,16,0)
+  option=GetImageOption(image_info,"heic:chroma-upsampling");
+  if (option != (char *) NULL)
+    {
+      if (LocaleCompare(option,"nearest-neighbor") == 0)
+        {
+          decode_options->color_conversion_options.
+            preferred_chroma_upsampling_algorithm =
+              heif_chroma_upsampling_nearest_neighbor;
+          decode_options->color_conversion_options.
+            only_use_preferred_chroma_algorithm = 1;
+        }
+      if (LocaleCompare(option,"bilinear") == 0)
+        {
+          decode_options->color_conversion_options.
+            preferred_chroma_upsampling_algorithm =
+              heif_chroma_upsampling_bilinear;
+          decode_options->color_conversion_options.
+            only_use_preferred_chroma_algorithm = 1;
+        }
+    }
+#endif
   if (preserve_orientation != MagickFalse)
     decode_options->ignore_transformations=1;
   chroma=heif_chroma_interleaved_RGB;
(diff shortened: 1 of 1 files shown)


coders/heic.c
```diff
@@ -1319,17 +1319,49 @@ static MagickBooleanType WriteHEICImage(const ImageInfo *image_info,
...
           {
-            error=heif_encoder_set_parameter(heif_encoder,"chroma",option);
-            status=IsHEIFSuccess(image,&error,exception);
-            if (status == MagickFalse)
-              break;
+            options->color_conversion_options.
+              preferred_chroma_downsampling_algorithm =
+                heif_chroma_downsampling_nearest_neighbor;
+            options->color_conversion_options.
+              only_use_preferred_chroma_algorithm = MagickTrue;
```
dlemstra, line 1345: I think we should use `1` here instead?

coders/heic.c
```diff
@@ -1319,17 +1319,49 @@ static MagickBooleanType WriteHEICImage(const ImageInfo *image_info,
...
+      {
+        error=heif_encoder_set_parameter(heif_encoder,"chroma",option);
+        status=IsHEIFSuccess(image,&error,exception);
+        if (status == MagickFalse)
+          break;
+      }
+#endif
+    options=heif_encoding_options_alloc();
+#if LIBHEIF_NUMERIC_VERSION >= 0x01100000
+    option=GetImageOption(image_info,"heic:chroma-downsampling-algorithm");
```
dlemstra, line 1336: Do we also want support for `preferred_chroma_upsampling_algorithm`? Also wondering if we should remove the `algorithm` word from the option.

dlemstra (commented):
Left some comments and I broke your PR because I added `HEIC_COMPUTE_NUMERIC_VERSION` that makes the versions checks a lot more readable.

coders/heic.c
```diff
@@ -1305,12 +1308,9 @@ static MagickBooleanType WriteHEICImage(const ImageInfo *image_info,
     status=IsHEIFSuccess(image,&error,exception);
     if (status == MagickFalse)
       break;
-#if LIBHEIF_NUMERIC_VERSION > 0x01060200
+#if LIBHEIF_NUMERIC_VERSION >= 0x01070000
```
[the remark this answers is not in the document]
njdoyle: This particular change is functionally equivalent. The libheif functionality we care about was introduced in version 1.7.0 but the check was "> 1.6.2". This lead me to some confusion initially. There are currently no versions between 1.6.2 and 1.7.0, and I don't expect there to be in the future, but I think this makes things more clear.
I think it should be safe to remove this check, though. I agree that `heif_encoder_set_parameter` should do the right thing here.

coders/heic.c
njdoyle: I'm happy to use `1` here. I wasn't sure what the proper style was here. libheif uses C `true` and `false` which I know ImageMagick avoids.

coders/heic.c
njdoyle: I'm personally indifferent about `preferred_chroma_upsampling_algorithm` because I don't expect to use it personally. I'm happy to add it though if you like.
I thought about removing the word `algorithm` too and left it in to mirror the words used by `libheif`. I'm not at all attached to keeping it though and wouldn't mind the shorter option name.

njdoyle:
@dlemstra, I've updated the PR to adopt your helpful macro and I think I've covered all of your other comments as well.

njdoyle:
Is there anything else I can do to help facilitate this getting merged?

dlemstra (approved):
Thanks for this PR! The delay was because we wanted to do the merge after the release of this weekend.

njdoyle:
I understand waiting for the next release to get this in. Thanks for the review and the merge!

PR #7194, opened 2024-03-25
Fix bmp option to bypass file size check

OnTheList-1:
### Prerequisites

### Description
ImageMagick provides an option for user to bypass file size check for bmp image. Currently, ImageMagick is retrieving option **after** file size check. This pull request aims to fix this by moving the option retrieval infront of the file size check, if the option 
_bmp:ignore-filesize_ is defined, actual file size won't be checked against file structure file size.


dlemstra (approved):
Thanks for this PR. We should check this in both places.

PR #7217, opened 2024-04-04
Updates to uhdr.c

aayushsoni111:
- handle images with odd dimensions. The library still does not support images with odd width / odd height.
- update configure.ac to locate installed uhdr library using pkg-config
- Improve exif metadata handling

Change-Id: I2e53dcc44493dcc017c0ffba98ab961aa106250b

### Prerequisites

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2892,6 +2892,9 @@ if test "$with_uhdr" != 'no'; then
     failed=0
     passed=0
 
+    PKG_CHECK_MODULES([UHDR],[libuhdr >= 1.0],[have_uhdr=yes],[have_uhdr=no])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -2892,6 +2892,9 @@ if test "$with_uhdr" != 'no'; then
     failed=0
     passed=0
 
+    PKG_CHECK_MODULES([UHDR],[libuhdr >= 1.0],[have_uhdr=yes],[have_uhdr=no])
```
kmilos, line 2895: See https://github.com/google/libultrahdr/issues/20#issuecomment-2036781883

kmilos: Looks like both the library name and pkgconf file settled on "libuhdr" upstream, so this is ok.

Issue #7262, opened 2024-04-19
Convert big PDF documents is slow


scresto09:
### Is your feature request related to a problem? Please describe.

Hello,

I noticed that saving large PDF files with ImageMagick was very slow, and it was generating the PDF ID entry which took a long time.
According to the PDF specification, this ID field is not required (page 47 of Document Management — Portable Document Format — Part 1: PDF 1.7).

### Describe the solution you'd like

I will add the **pdf:noID=true** option to the PDF encoder which allows this ID field not to be generated.

With a 3 MB JPG file, this goes from 1.12 s to 0.28 s.
![bench](https://github.com/ImageMagick/ImageMagick/assets/22740592/74aec3c9-34e0-44ba-9149-6892f99ba9c0)

dlemstra:
There really is no need to start a discussion, open an issue and also create a PR. Next time feel free to start with a pull request.

PR #7263, opened 2024-04-19
Convert big PDF documents is slow

scresto09:
### Prerequisites

### Description
I noticed that saving large PDF files with ImageMagick was very slow, and it was generating the PDF ID entry which took a long time.
According to the PDF specification, this ID field is not required (page 47 of Document Management — Portable Document Format — Part 1: PDF 1.7).
So I added the "pdf:noID" option to the PDF encoders which allows this ID field not to be generated.

With a 3 MB JPG file, this goes from 1.12 s to 0.28 s.

Fixes [#7262](https://github.com/ImageMagick/ImageMagick/issues/7262)


coders/pdf.c
```diff
@@ -3268,11 +3268,16 @@ static MagickBooleanType WritePDFImage(const ImageInfo *image_info,Image *image,
   (void) FormatLocaleString(buffer,MagickPathExtent,"/Root %.20g 0 R\n",(double)
     root_id);
   (void) WriteBlobString(image,buffer);
-  (void) SignatureImage(image,exception);
-  (void) FormatLocaleString(buffer,MagickPathExtent,"/ID [<%s> <%s>]\n",
-    GetImageProperty(image,"signature",exception),
-    GetImageProperty(image,"signature",exception));
-  (void) WriteBlobString(image,buffer);
+  option=GetImageOption(image_info,"pdf:noID");
```
dlemstra, line 3271: `pdf:no-identifier` would probably be a better name and more consistent with the rest of the names we use.

coders/pdf.c
```diff
@@ -3268,11 +3268,16 @@ static MagickBooleanType WritePDFImage(const ImageInfo *image_info,Image *image,
...
+  if (IsStringTrue(option) != MagickFalse) {
+    (void) DeleteImageProperty(image,"signature");
```
dlemstra, line 3273: There is no need to do this call?

coders/pdf.c
scresto09: OK, I renamed to pdf:no-identifier

coders/pdf.c
scresto09: OK, I removed this part.

scresto09:
Thanks for merge

PR #7593, opened 2024-09-05
add rsvg  to appimage

unclesamwk:
Add rsvg to appimage


dlemstra:
I finally had some time to review your PR. This does not add rsvg to the appimage but only updates our dev build to include this and we don't need that.

PR #7613, opened 2024-09-12
Don't overwrite LIBRARY_PATH in Makefile

### LINKED ISSUE #7611: Link failure due to cleared `LIBRARY_PATH`
### ImageMagick version

7.1.1.38

### Operating system

Linux

RHEL 8.7

### Description

When building with Pango/pangocairo the build fails when linking `MagickCore/.libs/libMagickCore-7.Q16HDRI.so.10.0.1` with
```
software/binutils/2.42-GCCcore-13.3.0/bin/ld: cannot find -lgobject-2.0: No such file or directory
software/binutils/2.42-GCCcore-13.3.0/bin/ld: cannot find -lglib-2.0: No such file or directory
software/binutils/2.42-GCCcore-13.3.0/bin/ld: cannot find -lharfbuzz: No such file or directory
software/binutils/2.42-GCCcore-13.3.0/bin/ld: cannot find -lcairo: No such file or directory
```

As a possible issue I found that in the Makefile `$LIBRARY_PATH` is overwritten by the install location of ImageMagick. This doesn't happen during configure so `pkg-config pangocairo` does not output paths present in `$LIBRARY_PATH`. So the `-L` flag is not returned and used in the Makefile and the linked can't find it in `$LIBRARY_PATH` which is mostly empty

### Steps to Reproduce

- Install Cairo to some custom location, e.g. `/opt/cairo`
- set LIBRARY_PATH to that (`/opt/cairo/lib`)
- configure and build

### Images

Build issue so no image

Flamefire:
The configure script sets this to a path inside the install location which is empty during build. However this path(s) is/are used by the compiler/linker to find libraries. `pkg-config` used during configure assumes this is the same during build as on invocation of itself and omits any path already contained in the variable.

This combination causes build failures due to dependent libraries not being found in the link step unless there happens to be some other way the paths get pulled in. E.g. if HarfBuzz is built using (the deprecated) configure&make it has `*.la` files which contain the required link flags. If HarfBuzz is built using the new Meson build system those files are no longer present and the linker paths will be missing.

As there doesn't seem to be a specific reason for the variable to be set to an empty directory in the Makefile just remove it.

Fixes #7611


urban-warrior:
`Makefile.in` is a generated file.  We instead need to patch `configure.ac` to resolve this issue.

Flamefire:
A yes, I though it is only `Makefile` which gets generated from `Makefile.in`. But I see this has been fixed in ba49d0cd543f6d0dd966be03bb09e185984e9635

I guess using `LIBRARY_ABSOLUTE_PATH` instead works, although I'd have suggested using  something with a clear suffix, such as `IM_LIBRARY_PATH`

Thanks for fixing!

PR #7734, opened 2024-11-01
.cut (Dr Halo) reading when run count in header

richard-smith-preservica:
### Prerequisites

### Description

When reading a .cut file where the run counts for a row are all together (RunCount > 0x80), the EncodedByte isn't decremented correctly.

By comparison with the working IM6 code (https://github.com/ImageMagick/ImageMagick6/blob/main/coders/cut.c#407), this is supposed to decrement by i+1, but the change to introduce the cast changed the precedence.

### DIFF
--- a/coders/cut.c
+++ b/coders/cut.c
@@ -384,7 +384,7 @@ static Image *ReadCUTImage(const ImageInfo *image_info,ExceptionInfo *exception)
       if (offset < 0)
         ThrowCUTReaderException(CorruptImageError,"ImproperImageHeader");
       if(EOFBlob(image) != MagickFalse) goto CUT_KO;  /*wrong data*/
-      EncodedByte=(size_t) ((ssize_t) EncodedByte-i+1);
+      EncodedByte=(size_t) ((ssize_t) EncodedByte-(i+1));
       ldblk+=(ssize_t) RunCountMasked;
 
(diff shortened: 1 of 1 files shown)


coders/cut.c
```diff
@@ -384,7 +384,7 @@ static Image *ReadCUTImage(const ImageInfo *image_info,ExceptionInfo *exception)
       if (offset < 0)
         ThrowCUTReaderException(CorruptImageError,"ImproperImageHeader");
       if(EOFBlob(image) != MagickFalse) goto CUT_KO;  /*wrong data*/
-      EncodedByte=(size_t) ((ssize_t) EncodedByte-i+1);
+      EncodedByte=(size_t) ((ssize_t) EncodedByte-(i+1));
```
dlemstra, line 387: Thanks for the fix but I am wondering if we could not write it like this instead:
```c
EncodedByte-=(size_t)(i+1);
```

richard-smith-preservica: I'm assuming the double cast was added for a good reason, so I left it there. The IM6 code is simply EncodedByte-=i+1 and I can't see how that could overflow, so if you want to change it then that would be fine too. (As far as I can tell, both EncodedByte and i are guaranteed to be small.)

dlemstra: Let's change it to `EncodedByte-=(size_t)(i+1);` for the readability.

PR #7773, opened 2024-11-14
Fix compiler identification with Clang on Darwin

Un1q32:
### Prerequisites

### Description

Moves the `__GNUC__` check below the `__clang__` check, as Clang defines `__GNUC__` when compiling for Darwin targets, identifying itself as GCC 4.2 for compatibility, resulting in imagemagick saying it was compiled with GCC 4.2 when it is compiled with Clang for Darwin.


dlemstra (approved):

dlemstra:

Un1q32:
Just found out this doesn't just affect Darwin, but Linux too, and probably other targets.

dlemstra:
It affects probably everyone building with clang. These changes are included in the latest release.

PR #7906, opened 2025-01-16
Magick++ Documentation Verification

akourieh:
## Description 
The current solution for the [documentation for Magick++](https://www.imagemagick.org/Magick++/Documentation.html) verification (discussed [here](https://github.com/ImageMagick/ImageMagick/discussions/7772#discussion-7479103) and here [#6704](https://github.com/ImageMagick/ImageMagick/discussions/6704)) involves a script that displays inconsistencies with function signatures between manually written documentation for Magick++ and Doxygen-generated documentation which is up to date. 

## GitHub Workflow
I made a GitHub Action that will run everyday at 6 AM like the other workflows there are. It will basically install the libraries needed and run a mini script. For the output of the mismatches, it is currently just printed out inside the workflow as described in the demo below.

In the workflow, there is a mini script that loops through all API documentation file (of Magick++) and the corresponding files generated by doxygen. It will run the Python script I made for each API file. Currently, not all files are supported. Indeed, I only search for the tables in the written HTML doc that contains something similar to "Method, Return, Signature, Description" since this is where the first issue arised from. Although, I realized some of the documentation also specifies class methods, attributes, etc. which is also doable in the future. These are the files in the Magick++ documentation that are not checked currently : `Drawable`, `STL`, `TypeMetric`.

## Python Script
The Python script is used to parse first through the documentation file and store all signatures (conform to the table as said above) in a dictionary. After cleaning all strings, I compare them with the Doxygen file generated previously to make sure that the documentation is still up to date with the current code. If there's a mismatch, it is printed out in the workflow.

In order to parse the HTML files accurately, I decided to make a Python script. However, I realized there are no Python scripts used anywhere in this project and I understand that in order to conform myself to this I should follow this pattern but I think it is just not efficient to do this in bash (which I tried previously). There are two very useful libraries in Python to do this parsing quite efficiently which are BeautifulSoup and Pandas.

I have tested the script and checked all files to see if the mismatches were all accurate and they were indeed. So currently, the scripts only displays the mismatches so anyone can go and change them manually. However, in the future, it would be possible to automate the update of the documentation (and I'm not sure if there is any need for a write permission ?).

## Demo of the workflow
So, if you go in the GitHub website of my forked repo https://github.com/akourieh/ImageMagick/ and in the Actions section, you can go on the latest run for the "Documentation Check for Magick++" and in there, if you look into the "Compare documentation" section, [you can see all the current mismatches printed](https://github.com/akourieh/ImageMagick/actions/runs/12807934078/job/35709567970), here is a sample of one below.

```
| Checking function signatures for 
|   Documentation file:  www/Magick++/Pixels.html
|   Doxygen file:  www/api/Magick++/classMagick_1_1Pixels.html
|
| - Return type mismatch for method "x"
| Doxygen return type   : ssize_t
| Documentation returns : ['int']
|
| - Return type mismatch for method "y"
| Doxygen return type   : ssize_t
| Documentation returns : ['int']
```
As we can see above, the return type for the "y" function is written to be "int" in the Magick++ API website, however it shows in the header of Pixels that it actually returns a "ssize_t".


dlemstra:
Thanks for your contribution. I will probably take a while before I have time to look at this.

.github/workflows/doc_check.yml
```diff
@@ -0,0 +1,42 @@
...
+    branches:
+      - main
+  workflow_dispatch:
+  schedule:
+    - cron: 0 6 * * *
+
+jobs:
+  doxygen-check:
+    name: 'Documentation Check'
+    runs-on: ubuntu-latest
```
dlemstra, line 14: Please change this to `ubuntu-24.04`.

.github/workflows/doc_check.yml
```diff
@@ -0,0 +1,42 @@
+name: Documentation Check for Magick++
+
+on:
+  push:
```
dlemstra, line 4: Please also add `paths:` and include this file so it will run every time we make a change to it. And do the same for `pull_request:` and include this file.

You can find more information about this in the docs or the `windows-release.yml` file.

.github/workflows/doc_check.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: |          
+          # first generate the Magick++ documentation to be up to date with the current code
+          cd config && doxygen Magick++.dox.in &> /dev/null
+          cd ..
+          
+          # documentation files with function signatures that we can compare with their corresponding doxygen file (currently)
+          DOC_HTML=("Blob.html" "CoderInfo.html" "Geometry.html" "Image++.html" "Montage.html" "Pixels.html")
+          DOXY_HTML=("Blob.html" "CoderInfo.html" "Geometry.html" "Image.html" "Montage.html" "Pixels.html")  # Image.html is the only one that differs from the doxygen file
+          for i in "${!DOC_HTML[@]}"; do
+            python3 scripts/compare_signatures.py www/Magick++/"${DOC_HTML[$i]}" www/api/Magick++/classMagick_1_1"${DOXY_HTML[$i]}"
```
dlemstra, line 40: I would like the build to fail when this script finds an issue in one of the files. But only after each file has been tested. And this doesn't seem to be happening now.

scripts/compare_signatures.py
```diff
@@ -0,0 +1,98 @@
+import re
```
dlemstra, line 1: Can you move this script and its folder to `.github/scripts/compare_signatures.py`?

dlemstra, line 1: Can change the output to print documentation and doxygen in the same order?

From
```
Checking function signatures for 
  Documentation file:  www/Magick++/Pixels.html 
  Doxygen file:  www/api/Magick++/classMagick_1_1Pixels.html 

- Return type mismatch for method "x"
Doxygen return type   : ssize_t
Documentation returns : ['int'] 
```

To
```
Checking function signatures for 
  Doxygen file       : www/api/Magick++/classMagick_1_1Pixels.html 
  Documentation file : www/Magick++/Pixels.html 

- Return type mismatch for method "x"
Doxygen return type   : ssize_t
Documentation returns : ['int'] 
```

dlemstra (commented):
I finally have made some time to take a look at your PR. Thanks for helping us with this. Gave it a test run and it looks promising! Can you please resolve my comments? If you don't have time for this I can also do this myself but I hope you can help us with this.

akourieh:
I have updated my branch with your comments and I have tested it this morning, sorry for the many commits I had issues running the workflow locally. Thank you a lot for the feedback !

dlemstra (commented):
Thanks for adding this for us.

dlemstra (approved):

Issue #7941, opened 2025-02-03
Crash when opening some exr files


mrsneeze:
### ImageMagick version

7.1.1-43

### Operating system

Windows

### Operating system, version and so on

10

### Description

Some EXR files will cause magick (amongst other things) to crash when they are decoded.

This looks like it's due to an issue in the underlying openexr library coming up with no channels, but there's an assumption that there is at least one channel in `exr.c` in the function `InitializeEXRChannels` where the channel name prefix is checked

### Steps to Reproduce

`magick Kapaa.exr out.jpg`

This will cause magick.exe to crash.

### Images

from: [openexr.com](https://openexr.com/en/latest/test_images/MultiResolution/Kapaa.html)

[Kapaa.zip](https://github.com/user-attachments/files/18641887/Kapaa.zip)

fmw42:
On my Mac OSX Ventura with IM 7.1.1.44 and OpenEXR 3.2.4,  I get

```
magick Kapaa.exr Kapaa.png
Abort trap: 6
```

I think IM only supports the half format (16-bits). Is this the full format?  An IM developer will need to comment. further.

I can open it in GraphicConverter on my Mac.  

`ExifTool says it is EXR Version: 2` 

```
New Features for OpenEXR 2.0: Multi-Part and Deep Data

The multi-part format is an extension of the OpenEXR 1.7 single-part file format. In addition to supporting the OpenEXR 1.7 data storage (a single scan line or tiled image), OpenEXR 2.0 files can be used to store multiple views and/or deep data (deep scan line or deep tiles).

```

**So it looks like there might be a new version that IM has not updated.**


```
---- ExifTool ----
ExifTool Version Number:
12.65
---- System ----
File Name:
Kapaa.exr
Directory:
/Users/fred/Desktop
File Size:
6.8 MB
File Modification Date/Time:
2025:02:03 11:56:20-08:00
File Access Date/Time:
2025:02:03 08:09:47-08:00
File Inode Change Date/Time:
2025:02:03 08:09:47-08:00
File Permissions:
-rw-rw-r--
---- File ----
File Type:
EXR
File Type Extension:
exr
MIME Type:
image/x-exr
Image Width:
799
Image Height:
546
---- OpenEXR ----
EXR Version:
2
Flags:
Tiled
Channels:
B half 1 1, G half 1 1, R half 1 1
Compression:
ZIP
Data Window:
0 0 798 545
Display Window:
0 0 798 545
Exposure Time:
[...]

PR #7942, opened 2025-02-03
#7941 Added if statement to avoid crash when there are no channels

mrsneeze:
### Prerequisites

### Description

For #7941 there is an assumption that at least one channel is found while decoding as the first entry in the array is used when checking the prefix.

This change checks that we do have at least one channel before we use the channels array.


dlemstra:
Thanks for your PR but we will not merge it. I solved this differently, the problem was at a different spot.

PR #8196, opened 2025-06-12
Create Install-OpenHarmony.txt for Cross-Compile method of OpenHarmony

kuniqi:
we are using ImageMagick in OpenHarmony, so we'd like to contribute the build guide, hope it can help developers.


dlemstra:
I don't think we want to add this file to our repository.

PR #8230, opened 2025-06-30
Fix NULL pointer dereference in XWarning

moon044:
### Prerequisites

### Description

Hi,
This PR addresses a potential NULL pointer dereference (CWE-476) in `MagickCore/xwindow.c`.

The result of `XSetWindows((XWindows *) ~0)` is used without checking for `NULL`.  
Also, the `description` parameter is passed directly into `XNoticeWidget()` without validation.

If either pointer is `NULL`, it could lead to a segmentation fault or undefined behavior.

---

#### Expected behavior

- All pointers used in function calls must be validated before dereferencing.
- Specifically, `windows` and `description` should be checked for `NULL`.

#### Actual behavior

- No `NULL` check exists before dereferencing `windows->display`.
- No check for `description` being `NULL`, which is cast and used directly.
- This may result in a segmentation fault under edge conditions.

### Patch

This patch adds minimal but sufficient checks before the call to `XNoticeWidget()`:

```diff
windows = XSetWindows((XWindows *) ~0);
+ if (windows == NULL || description == NULL)
+     return;
XNoticeWidget(windows->display, windows, text, (char *) description);
```

---

### Notes
This patch introduces only minimal changes to address the specific issue, without altering the overall behavior of the function.
This change does not alter existing behavior except to prevent a potential crash (segmentation fault).

Thank you for your consideration. I look forward to your feedback.

### DIFF
--- a/MagickCore/xwindow.c
+++ b/MagickCore/xwindow.c
@@ -9708,7 +9708,8 @@ MagickPrivate void XWarning(const ExceptionType magick_unused(warning),
   (void) CopyMagickString(text,reason,MagickPathExtent);
   (void) ConcatenateMagickString(text,":",MagickPathExtent);
   windows=XSetWindows((XWindows *) ~0);
-  XNoticeWidget(windows->display,windows,text,(char *) description);
+  if (windows != (XWindows *) NULL)
+     XNoticeWidget(windows->display,windows,text,(char *) description);
 }
 
 /*


MagickCore/xwindow.c
```diff
@@ -9708,6 +9708,8 @@ MagickPrivate void XWarning(const ExceptionType magick_unused(warning),
   (void) CopyMagickString(text,reason,MagickPathExtent);
   (void) ConcatenateMagickString(text,":",MagickPathExtent);
   windows=XSetWindows((XWindows *) ~0);
+  if (windows == NULL || description == NULL)
```
dlemstra, line 9711: Please remove the `description  == NULL` check because this is already done inside the method. And please change your patch to:

```c
if (windows != ((XWindows *) NULL)
   XNoticeWidget(windows->display,windows,text,(char *) description);
```

dlemstra (commented):
Thanks for the fix. Even though I don't think we ever reach this point when `XSetWindows` would return `NULL` it is probably still good to add a check. Can you update your patch to what I suggested in my comment?

dlemstra (approved):
Thanks for this!

PR #8243, opened 2025-07-06
Add support for Simple File Format Family (SF3) images

Shinmera:
### Prerequisites

### Description
This pull request aims to add support for the Simple File Format Family (SF3) image file format as described here: https://shirakumo.github.io/sf3/specification#image with sample files available here: https://shirakumo.org/projects/sf3/tree/master/samples/

I've tried my best to figure things out on my own and the encoder and decoder are working properly. I can convert files, display them, the works. There are a few things that I'm unsure about though:

I couldn't find a clear description of what the magick coder flags are all supposed to mean exactly. Namely:
- `CoderDecoderSeekableStreamFlag` `CoderEncoderSeekableStreamFlag`  
  Are these meant to signify that a seekable stream is *supported*, or that a seekable stream is *required*? In the case of SF3, a seekable stream is not mandatory for decoding, and optional for encoding (to back-fill the CRC32 checksum in the header).
- `CoderRawSupportFlag`  
  Is this meant to indicate that the coder reads raw, uncompressed image data? Or image data outside of float range? or?... SF3 is uncompressed and does support floating point pixel data.
- `CoderBlobSupportFlag` `CoderStealthFlag` `CoderUseExtensionFlag`
  I have honestly no clue what these flags indicate and haven't been able to tell from other coders, either.

For whatever reason when invoking `magick.sh convert a.png a.sf3`, it encodes to PNG instead of SF3. The encoder works fine when invoked as `convert a.png sf3:a.sf3`. As far as I could tell from grepping through the source for other coders, I've done all the necessary hookup bits including entries in the magic.xml database. I would very much appreciate if someone could tell me what I've done wrong there!

If there's anything else I should change or fix, please let me know.


dlemstra:
Thank you for this pull request but we are not preparing for a new release and we don't want to include this PR in those changes. We will take a look at your pull request after the next release and then decide what we want to do with it.

fmw42:
Please explain the benefits of this file format.  Why do you need it and why can you not use some other file format?

Shinmera:
SF3 file formats are aimed to be simple to implement and particularly support workflows where the data can be accessed directly in memory (via mmap/etc) without any extra decoding work, making them very fast to access. This can be useful in contexts like game engines.

Even aside from that particular context, there is value in supporting a plain bitmap format that is as simple as possible and still well specified in an independent standard. Existing formats for bitmaps like BMP and TGA nevertheless include a lot of complexity that makes them non-trivial to implement and support.

fmw42:
NetPBM is a very simple format without a lot of header stuff

Shinmera:
As far as I understand NetPBM does not support things such as floating point channels or layers among other things.

Shinmera:
Thank you very much for the merge, @urban-warrior!

To sate my curiosity, is what I mentioned at the end of my note an artefact of my setup being wrong somehow?

> For whatever reason when invoking magick.sh convert a.png a.sf3, it encodes to PNG instead of SF3. The encoder works fine when invoked as convert a.png sf3:a.sf3. As far as I could tell from grepping through the source for other coders, I've done all the necessary hookup bits including entries in the magic.xml database. I would very much appreciate if someone could tell me what I've done wrong there!

urban-warrior:
Our preference is to force the user to specify the format explicitly for a new untested format.  There may be unanticipated security issues with this image format.  After the format proves itself stable and secure we can revisit honoring the implicit format, e.g., image.sf3.

Shinmera:
Ah, that's fair enough, thanks for answering!

I noticed that the CI is failing to find `zlib.h`. Many apologies for that, I followed the include from `png.c` and I didn't test thoroughly enough with the various build flags. I'm working on a patch to address that.

Shinmera:
Okey, the patch at shinmera/imagemagick@c61b682 should remove the dependency on zlib using a static crc32 implementation, if that's preferred over linking against zlib for sf3.

[0001-Remove-dependency-on-zlib-for-sf3.patch.txt](https://github.com/user-attachments/files/21205429/0001-Remove-dependency-on-zlib-for-sf3.patch.txt)

Again, many apologies for the trouble I've caused!

Shinmera:
Similarly, Shinmera/ImageMagick@b19bea22e9a94d3864c0d0bebf2f0d82521d5dcb should fix the -Werror=implicit-fallthrough that the CI complains about. With that the compile gets through sf3 on my system (though complains in exr.c)

[0001-Minor-change-to-fix-complaint-arising-from-Werror-im.patch.txt](https://github.com/user-attachments/files/21205540/0001-Minor-change-to-fix-complaint-arising-from-Werror-im.patch.txt)

urban-warrior:
Check `tests/validate` as the SF3 validation is failing.  We disabled the validation for now.

PR #8383, opened 2025-10-04
Allow raw PICT files (no padding nor magic) to be read (#8381)

### LINKED ISSUE #8381: Cannot read old Quickdraw PICT without a 512 bytes padding
### ImageMagick version

7.1.1-47

### Operating system

Other (enter below)

FreeBSD 14.3-RELEASE

### Description

An old PICT of mine from 2001, embedded in an `.rtfd` (probably generated on ClarisWorks / AppleWorks), cannot be converted by ImageMagick because:
> magick: improper image header `invit2.pict' @ error/pict.c/ReadPICTImage/918.

`hexdump`ing the PICT file:
```
00000000  6b 44 00 00 00 00 00 a2  01 0e 11 01 01 00 0a 00
```
shows the version 1 PICT header at offset 10, that is, without the "usual" 0-filled 512 bytes pre-header;
I could find a mention to this alternate layout on http://justsolve.archiveteam.org/wiki/PICT#Identification, from what I understand this padding is a free place for application use.

And indeed, the [corresponding source code](https://github.com/ImageMagick/ImageMagick/blob/main/coders/pict.c#L911-L915) shows that the 512 bytes are skipped unconditionally (unless there are the `PICT` magic bytes).

Handling all variants (with or without magic `PICT`, with or without 512 bytes padding) would probably require a more thorough inspection of possible magic bytes before deciding to skip.

### Steps to Reproduce

```sh
magick invit1.pict invit1.jpg
magick invit2.pict invit2.jpg
```

### Images

https://tmp.outters.eu/invit1.pict (version 2 PICT)
https://tmp.outters.eu/invit2.pict (version 1 PICT)

outtersg:
### Prerequisites

### Description

Handle raw Quickdraw PICT files in addition to the OLE 2-prefixed ones (having 4 magic bytes `PICT`) and 512-bytes padded ones (bytes available for application use).
This resolves #8381, where examples and pointers to (retro-engineered?) "specifications" are given.

### Way of solving

We have to read the first 4 bytes to tell if they represent the OLE magic bytes "`PICT`"; in case they do not match, we shall re-read them (and the following 6 bytes) as 2 bytes, then 8 bytes, that _may_ hold the "raw" layout (2 bytes for image size, 8 bytes for its frame).
 
The general way would have been to rewind the `Blob` when detecting that 4 already bytes aren't what we were looking for, but rewinding is not possible on streams.
So I applied what looked to me the most natural way to implement it within ImageMagick's toolbox, but as I am very new to ImageMagick, I can't pretend this is the better or even right way to act (so _if I'm wrong feel free to give me pointers_):
* `CloneBlobInfo(NULL)` and `AttachBlob()` do temporarily replace the image's Blob with an in-memory one, composed from the 4 bytes + 6 others read from the real Blob
* the `ReadRectangle()` call can then be done on this image, given that _none of the underlying calls to `ReadBlob()` will tamper the `Image`_  (thus ensuring we have no side effects from the temporary `Blob` to its host `Image`)
* then we can swap back the real `Blob` to the `Image`, and continue reading

To avoid a lot of `goto`s I put all the code in a new `ReadHeaderUpToVersion()` (with `return(MagickFalse)` for quick return as soon as we've seen a problem), which helps:
* making the whole process clear (try to find the header with possibly its magic `PICT` bytes, if unsuccessful then skip the remaining padding bytes up to the 512th, and retry finding a header from there)
* reducing the number of `ThrowPICTException(CorruptImageError,"ImproperImageHeader")`

Tested with the 3 variants (raw, 4-bytes magic, 512-bytes padding) of my 2 images (one of version 1 PICT, one of version 2 PICT):
```sh
for f in invit[12].pict
do
    for header in true "printf PICT" "dd if=/dev/random bs=512 count=1 status=none"
    do
        { $header ; cat $f ; } > 1.pict
        echo "=== $f / $header ==="
        > 1.jpeg # Make an empty file so that in case of convert error the ending identify will complain.
        magick identify 1.pict
        magick 1.pict 1.jpeg
        magick identify 1.jpeg
    done
done
```

### DIFF
--- a/coders/pict.c
+++ b/coders/pict.c
@@ -816,6 +816,101 @@ static MagickBooleanType ReadRectangle(Image *image,PICTRectangle *rectangle)
   return(MagickTrue);
 }
 
+static MagickBooleanType ReadHeaderUpToVersion(Image *image,
+  PICTRectangle *frame,ssize_t *version)
+{
+  BlobInfo
+    *starter_blob,
+    *original_blob;
+
+  char
+    header[2 + sizeof(PICTRectangle)];
+
+  int
+    c;
+
+  MagickBooleanType
+    success;
+
+  ssize_t
+    i;
+
+  original_blob=(BlobInfo *) NULL;
+
+  header[0]=(char) ReadBlobByte(image);
+  header[1]=(char) ReadBlobByte(image);
+  header[2]=(char) ReadBlobByte(image);
+  header[3]=(char) ReadBlobByte(image);
+  /*
+    If our first 4 bytes were not PICT, we have to rewind and retry reading
+    them as size and frame. As we cannot rewind image, we use an intermediate
+    Blob where we aggregate already read bytes with remaining ones.
+  */
+  if (!((header[0] == 0x50) && (header[1] == 0x49) &&
+      (header[2] == 0x43) && (header[3] == 0x54 )))
+    {
+      for (i=4; i < 2+sizeof(PICTRectangle); ++i)
+        if ((c = ReadBlobByte(image)) == EOF)
+          break;
+        else
+          header[i] = (char) c;
+      starter_blob = CloneBlobInfo(NULL);
+      if (starter_blob==NULL)
+        return(MagickFalse);
+      AttachBlob(starter_blob, header, i);
+      original_blob = image->blob;
+      image->blob = starter_blob;
+    }
+  (void) ReadBlobMSBShort(image);  /* skip picture size */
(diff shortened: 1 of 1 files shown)


dlemstra:
Thanks for your pull request but I decided to solve this differently: 9002a93a94ff151795624499da3535a0c9408d1a. I can read both files from the issue but one of them is completely black. So that might be another issue.

Issue #8425, opened 2025-11-03
pgm P2 Format Parsing Error When Final Line Lacks Newline Character


zhao-leo:
### ImageMagick version

7.1.1-43

### Operating system

Linux

### Operating system, version and so on

Linux/ubuntu 6.14.0-35-generic 

### Description

When reading a pgm file in P2 format, the parser fails if the last line of pixel data does not end with a newline character.

Instead of throwing an error, I guess the parser may handle this case gracefully and process the data correctly.

I believe providing parsing functionality instead of throwing errors can effectively enhance stability. In tests with MATLAB, it was able to read such files without issues.

### Steps to Reproduce

Input:
```bash
magick 256x256_p2.pgm output.png
```
Result:
```
magick: unable to read image data `256x256_p2.pgm' @ error/pnm.c/ReadPNMImage/1591.
```

### Images

[images.zip](https://github.com/user-attachments/files/23314876/images.zip)

dlemstra:
Thanks for reporting this and creating a pull request to solve this. I took some inspiration from your pull request but decided to solve it differently. Support for reading those two files that you provided will become available in the next release.

PR #8426, opened 2025-11-04
fix(coder/pnm.c):Detect and handle premature EOF in PNM reader

zhao-leo:
### Prerequisites

#### BUG
1. When P2 format PGM files end without a trailing newline after the last pixel value, the `PNMInteger` function correctly reads the final digit
2. However, when it attempts to read the next character to determine if the number has ended, it encounters EOF
3. The original code would immediately report "UnableToReadImageData" error, even though all required pixel data had been correctly read

#### Fix
Mainly add pixel counting logic for P2 formats and add a flag marker.

I tested the code on my computer using normal p2-pgm, the image that caused the error and the damaged image, and the program performed correctly.

### DIFF
--- a/coders/pnm.c
+++ b/coders/pnm.c
@@ -222,7 +222,7 @@ static unsigned int PNMInteger(Image *image,CommentInfo *comment_info,
       }
     c=ReadBlobByte(image);
     if (c == EOF)
-      return(0);
+      return(value);
   }
   if (c == (int) '#')
     c=PNMComment(image,comment_info,exception);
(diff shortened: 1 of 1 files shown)


zhao-leo:
Demo images can be downloaded from #8425 .

dlemstra:
Thanks for your pull request but I decided to solve this slightly different: 7404bd00879aa0a515d7760d6fdd889e39542f41. Without your PR I would have probably missed the `return(value)` change so I would still like to thank you for creating this PR!

PR #8445, opened 2025-11-14
JPEG: Support 2-to-16-bit lossless JPEG images with libjpeg-turbo 3.1+; Support additional IDCT scaling factors with libjpeg-turbo and libjpeg v7+

dcommander:
### Prerequisites

### Description
- Add support for 3/8, 5/8, 3/4, 7/8, 9/8, 5/4, 11/8, 3/2, 13/8, 7/4, 15/8, and 2/1 scaling factors if IM is built against libjpeg-turbo or libjpeg v7+.
- Add support for any legal lossless data precision (from 2 to 16 bits per sample inclusive) if IM is built against libjpeg-turbo 3.1+.  This also improves the accuracy of converting samples between different data precisions in the JPEG module.
- Add support for generating 12-bit-per-sample lossy JPEG images if IM is built against libjpeg-turbo 3.0+.

These changes have been validated using [a set of scripts](https://github.com/dcommander/imtest) that iterates over the various permutations of JPEG settings and formats using separate builds of IM targetting libjpeg-turbo 3.1.x, 3.0.x, and 2.1.x, as well as libjpeg v6b with Ken Murchison's lossless JPEG patch.

### DIFF
--- a/coders/jpeg.c
+++ b/coders/jpeg.c
@@ -256,12 +256,23 @@ static MagickBooleanType IsJPEG(const unsigned char *magick,const size_t length)
 %
 */
 
-static inline int JPEGGetSample(const struct jpeg_decompress_struct *jpeg_info,
-  const JSAMPLE *p)
+static inline Quantum JPEGGetQuantum(
+  const struct jpeg_decompress_struct *jpeg_info,const JSAMPLE *p)
 {
-  if (jpeg_info->data_precision > 8)
-    return((int) (*(unsigned short *) p));
-  return((int) (*(JSAMPLE *) p));
+  if (jpeg_info->data_precision == 8)
+    return ScaleCharToQuantum(*p);
+  else if (jpeg_info->data_precision == 16)
+    return ScaleShortToQuantum(*(unsigned short *) p);
+  else if (jpeg_info->data_precision < 8)
+    {
+      QuantumAny range=GetQuantumRange(jpeg_info->data_precision);
+      return ScaleAnyToQuantum(*p,range);
+    }
+  else
+    {
+      QuantumAny range=GetQuantumRange(jpeg_info->data_precision);
+      return ScaleAnyToQuantum(*(unsigned short *) p,range);
+    }
 }
 
 static boolean FillInputBuffer(j_decompress_ptr compress_info)
@@ -1093,7 +1104,6 @@ static Image *ReadOneJPEGImage(const ImageInfo *image_info,
 
   ssize_t
     i,
-    scale,
     y;
 
   struct jpeg_error_mgr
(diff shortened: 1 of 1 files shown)


dcommander:
Force-pushed to correct a couple of formatting issues and fix the Windows build.

dlemstra:
In the next release writing high depth jpeg files will require setting the `jpeg:high-bit-depth` define. This will not be enabled by default anymore because a lot of decoders will not support reading these kind of files.

dcommander:
Playing Devil's advocate, can't users simply force the output bit depth to 8 if they want compatibility with more limited decoders?  If you convert, say, a 16-bit PPM image into a 16-bit TIFF or PNG image, ImageMagick doesn't silently drop the bit depth in order to preserve backward compatibility with more limited decoders.  It fully uses the capabilities of the most popular TIFF and PNG decoders.  libjpeg-turbo is the most popular JPEG decoder, as well as the official ISO/ITU-T reference implementation.

My main concern is that there may be a chicken-and-egg issue.  If ImageMagick supports all JPEG-1 variants by default, then it will encourage other applications to do likewise, and those other applications can now easily do so by using libjpeg-turbo.  However, if ImageMagick doesn't support all JPEG-1 variants by default, then other applications will have little impetus to do so, and higher-bit-depth and lossless images will remain orphaned even though they are conformant JPEG-1 images.

Perhaps disabling `jpeg:high-bit-depth` by default could be temporary, until libjpeg-turbo 3.0.x and later become more widely available in operating systems?  Or perhaps it could be enabled by default only if ImageMagick is built with libjpeg-turbo 3.1.x and later?

A bit of historical context:
libjpeg originated in an era before shared libraries were common, so it was designed to be compiled and linked statically with individual applications.  Thus, it supports 12-bit lossy JPEG images, but only if you recompile it in a way that collides with the namespace of, and breaks API/ABI compatibility with, the 8-bit version of the library.  Thus, the default feature set of libjpeg became canonical once the library was integrated into Linux and other operating systems.  That limited the adoption of 12-bit lossy JPEG images, even though technically they have been supported all along.  Similarly, Tom Lane stepped away from the IJG before he could integrate lossless support, so people who wanted to encode/decode lossless JPEG images had to compile libjpeg with Ken Murchison's patch that also collides and breaks compatibility with the 8-bit lossy version of the library.  That limited the adoption of lossless JPEG images.  libjpeg-turbo attempts to address those historical limitations by providing support for all JPEG-1 variants using the same API/ABI.

arekm:
Even common browsers (firefox, google chrome) don't support 12-bit jpeg and I just found out about that (after Image Magick upgrade (to 7.1.2-9) which "broke" thumbnail generation on some web pages. Now on 7.1.2-13, so problem solved here).

PR #8456, opened 2025-11-23
Add an alias for the original Magic Kernel

warningnonpotablewater:
### Prerequisites

### Description

ImageMagick currently includes Magic Kernel Sharp 2013 and Magic Kernel Sharp 2021, but not the original (non-sharp) Magic Kernel from 2011. Luckily, `Quadratic` happens to be the exact same kernel:

Analytical formula for the Magic Kernel, identical to Quadratic

```c
static double Quadratic(const double x,
  const ResizeFilter *magick_unused(resize_filter))
{
  magick_unreferenced(resize_filter);

  /*
    2rd order (quadratic) B-Spline approximation of Gaussian.
  */
  if (x < 0.5)
    return(0.75-x*x);
  if (x < 1.5)
    return(0.5*(x-1.5)*(x-1.5));
  return(0.0);
}
```

This PR adds a `MagicKernel2011` alias for completeness.


dlemstra:
Why would we need this alias? Can you not use quadratic?

warningnonpotablewater:
Well... You're right. I just thought this was a gap in names worth closing. Sorry for the disturbance.

Issue #1577, opened 2019-05-23
Cannot convert png to ico larger than 255x255


maxloh:
### Prerequisites

- [o] I have written a descriptive issue title
- [o] I have verified that I am using the latest version of ImageMagick
- [o] I have searched [open](https://github.com/ImageMagick/ImageMagick/issues) and [closed](https://github.com/ImageMagick/ImageMagick/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

ImageMagick could only generate valid ico file smaller than or equal to 255x255
It fails to generate valid 256x256 ico that can be correctly display by Windows

### Steps to Reproduce

- Download `sample.png`
![sample](https://user-images.githubusercontent.com/35831069/58270795-326ff600-7dbd-11e9-99de-e88ffcaf2f89.png)
- Run `magick convert sample.png -define icon:auto-resize=255 255.ico`, the generated 255x255 ico file can be displayed by Windows (See attached 255.ico file)
- Run `magick convert sample.png -define icon:auto-resize=256 256.ico`, the generated 255x255 ico file cannot be displayed by Windows (See attached 256.ico file)

### System Configuration

- ImageMagick version: 

ImageMagick 7.0.8-46 Q16 x64 2019-05-19
- Environment (Operating system, version and so on): Windows 10 Pro 18362.116
- Additional information:

fmw42:
This may not help, but in IM 7, use magick, not magick convert. The latter may use IM 6 code.

maxloh:
@fmw42 I tried to use magick only but the file is invalid too.
```magick sample.png -define icon:auto-resize=256 256-no-convert.ico```
256-no-convert.ico https://drive.google.com/open?id=1bZ3j7Pvt-TeHwtiEaV0AovCT_fCbJPLq

fmw42:
Yes, I thought it would not matter. But suggested it just in case. Sounds like a proper bug.

urban-warrior:
Icons with a width or height of 256 are stored as PNG blobs.  Windows supports this but not all Windows clients can recognize this format.

maxloh:
@urban-warrior Is it possible to store 256x256 ico as BMP blobs with ImageMagick, like what we got when we convert image to ico smaller than 256x256?

heyakyra:
Not only is it a proper bug but it was supposed to be fixed in version 6.8.9-8 according to this forum thread back in 2014:
[...]

PR #8569, opened 2026-02-15
ICON: allow writing 256x256 icon rasters

Tachi107:
The previous code would limit itself to writing 255x255 images, while ICON supports rasters up to 256 pixels; to do so, the width and/or height have to be 0, as that is interpreted as 256.

Actually fixes https://github.com/ImageMagick/ImageMagick/issues/1577

### DIFF
--- a/coders/icon.c
+++ b/coders/icon.c
@@ -1100,7 +1100,7 @@ static MagickBooleanType WriteICONImage(const ImageInfo *image_info,
       bits_per_pixel,
       planes;
 
-    if ((next->columns > 255L) && (next->rows > 255L) &&
+    if ((next->columns > 256L) && (next->rows > 256L) &&
         ((next->compression == UndefinedCompression) ||
         (next->compression == ZipCompression)))
       {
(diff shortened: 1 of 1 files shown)


STPR:
Hello,
Can you please revert this ?

With this patch, contrary to what is usually done (compress the sizes equal to 256 or above), the 256x256 is now **not converted** to PNG.
Even if `-compress zip` is specified.

A better solution could be to NOT include PNG when the `-compress zip` parameter  is NOT specified.
And let the the sizes equal to 256 and above treated as PNG only when the `-compress zip` parameter is specified.

STPR:
Have you tested with sizes of 512 and 1024 with your %256 ???

Tachi107:
I’m confused. Why should 256x images be PNG *by default*? If I’m asking for a ICO, I want a Microsoft ICON file. I do agree that it should be possible to create ICOs with embedded PNGs, but 256 should not be a magic number triggering PNG embedding. And, really, it should also be possible to create PNG-compressed ICOs with sizes lower than 256x.

Reverting the patch I proposed would not fix the issue. A new patch, actually honouring command line flags specifying compression, should be written.

STPR:
https://en.wikipedia.org/wiki/ICO_(file_format)
`Microsoft recommends that all 256×256 icons in ICO files should be stored in PNG format to reduce the overall size of the file.`
You can check a lot of ico files. It's usually done this way.
I guess it's a compromise between size and the CPU it takes to unzip the PNG. Even if it's probably less true nowadays.

The problem with this patch, while it doesn't solve the problem, makes impossible the ico files creation the way it's usually done.
My .ico files are now 3 times bigger...
Even when I use the `-compress zip` parameter as it just doesn't compress the 256x256 size anymore and store it as an ICO file.

Tachi107:
> The problem with this patch, while it doesn't solve the problem, makes 
> impossible the ico files creation the way it's usually done.

What do you mean? It does solve the problem. The issue you are 
describing was already present before this patch, as ImageMagick would 
ignore the compression parameter for all sizes below or equal to 
255x255. It now does for 256x256 images as well, but it's just a side 
effect of fixing a different bug.

Please stop spamming and consider opening a new issue tracking the 
impossibility of applying `-compress zip` to all ICO files of 256x256 
pixels. If you have more time, you might consider working on a patch as 
well; having looked at the code, it doesn't seem that hard.

PR #8600, opened 2026-03-04
configure: add `--with-pkgconfig-static` flag to support static linking to pkg-config sourced libraries

juchem:


dlemstra:
There is no need to add this? Static builds are already supported.

juchem:
@dlemstra if all I have is .a libraries and no .so, then imagemagick build fails at link time with a bunch of undefined references.

The reason is transitive dependencies, they're not linked against when pkgconf is not given `--static`.

PR #8606, opened 2026-03-09
Amiga Workbench .info coder (image/x-amiga-icon)

bitplane:
### Prerequisites

### Description

This PR implements a coder for reading Amiga Workbench .info files, containing icons used by applications of the day. These come in 5 flavours:

1. Classic Workbench 1.x to 3.x, using the system's palette.
2. NewIcons - mid-90s hack to encode pixels in ToolTypes
3. GlowIcons - RLE encoded IFF images
4. ARGB - zlib compressed pixel data
5. PNG - A pair of PNG files concatenated with extra metadata in the first one

The coder supports the first 4 types of icon, the 5th form is extremely rare and I only found a few examples.

#### Test data

It's been tested against ~300,000 .icon files mined from Aminet ([available here](https://github.com/bitplane/amiga-info-test-files)), and handles the garbled ones gracefully.

### Piccy

They say a picture is worth 0x1000 DWORDS. Here's a bunch of files converted with `magick convert` - showing the messy state of backwards compatibility:

image


urban-warrior:
It looks like you are making improvements to this pull request.  We'll sit on a review for a week to give you ample time to stabilize the coder module.

bitplane:
Hi, any time to review this? I don't think it needs any more work.

coders/coders.h
```diff
@@ -20,6 +20,7 @@
   Include declarations.
 */
 #include "coders/aai.h"
+#include "coders/wbinfo.h"
```
dlemstra: This is ordered alphabetically.

coders/Makefile.am
```diff
@@ -114,6 +114,8 @@ MAGICK_CODER_CPPFLAGS = \
 MAGICKCORE_CODER_SRCS = \
 	coders/aai.c \
 	coders/aai.h \
+	coders/wbinfo.c \
```
dlemstra: Here also and in other places

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
+/*
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                             %
+%                                                                             %
+%                                                                             %
+%     W   W  BBBB   IIIII  N   N  FFFFF   OOO                                 %
```
dlemstra: Please center this.

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+#include "MagickCore/memory_.h"
+#include "MagickCore/monitor.h"
+#include "MagickCore/monitor-private.h"
+#include "MagickCore/pixel-accessor.h"
+#include "MagickCore/quantum-private.h"
+#include "MagickCore/static.h"
+#include "MagickCore/string_.h"
+#include "MagickCore/module.h"
+#if defined(MAGICKCORE_ZLIB_DELEGATE)
+#include <zlib.h>
```
dlemstra, line 62: This should also be linked in the Makefile.am file (see psd coder for example)

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+};
+
+static const unsigned char AmigaWB2x[8][3] =
+{
+  { 149, 149, 149 }, { 0, 0, 0 }, { 255, 255, 255 }, { 59, 103, 162 },
+  { 123, 123, 123 }, { 175, 175, 175 }, { 170, 144, 124 }, { 255, 169, 151 }
+};
+
+/*
+  Helper: extract an unsigned integer from a bit array.
```
dlemstra: Not sure if we need all these captain obvious comments.

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+*/
+static unsigned int BitsToUInt(const unsigned char *data, size_t total_bits,
```
dlemstra: Our code style is not to put spaces between the arguments.

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+
+    byte=(unsigned int) (unsigned char) src[ch_idx];
+    if (byte < 0x20)
+      continue;
+    if (byte <= 0x9F)
+      val=byte-0x20;
+    else if (byte <= 0xD0)
+      val=byte-0x51;
+    else if (byte >= 0xD1)
+    {
```
dlemstra: Our code style is a bit different:

```C
else if
  {
  }
```

p.s. I won't comment every spot.

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+  {
+    unsigned int
+      control;
+
+    control=BitsToUInt(data,total_bits,&bit_pos,8);
+    if (control == 0x80)
+      continue;
+    if (control <= 0x7F)
+    {
+      size_t
```
dlemstra: You need this in both places. Can you move this up to improve the readability?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+  count=0;
+  while ((bit_pos+8 <= total_bits) && (count < expected_count))
+  {
+    unsigned int
+      control;
+
+    control=BitsToUInt(data,total_bits,&bit_pos,8);
+    if (control == 0x80)
+      continue;
+    if (control <= 0x7F)
```
dlemstra: `127` would make this more readable, same for `0x80`.

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+  {
+    AcquireNextImage(image_info,*image,exception);
+    *image=SyncNextImageInList(*image);
+    if (*image == (Image *) NULL)
+      return(MagickFalse);
+  }
+  if (SetImageExtent(*image,width,height,exception) == MagickFalse)
+  {
+    if (scene > 0)
+      DeleteImageFromList(image);
```
dlemstra: Why are you doing this?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+    {
+      size_t
+        pi;
+
+      unsigned char
+        idx;
+
+      pi=(size_t) y*width+(size_t) x;
+      idx=pi < pixel_count ? pixels[pi] : 0;
+      if (has_transparent && (idx == transparent_idx))
```
dlemstra, line 383: I think this would be more readable:

```c
 if ((palette != (const unsigned char *) NULL) &&
     ((size_t) idx < pal_entries))
  {
    // SetPixelRed/Green/Blue/Alpa
  }
else
  {
	SetPixelRed(image,0,q);
    SetPixelGreen(image,0,q);
    SetPixelBlue(image,0,q);
    if (has_transparent && (idx == transparent_idx)
      SetPixelAlpa(image,TransparentAlpha,q);
    else
      SetPixelAlpa(image,OpaqueAlpha,q);
  }
```

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+  const unsigned char
+    (*palette)[3];
+
+  MagickBooleanType
+    render_status;
+
+  ssize_t
+    y;
+
+  (void) ReadBlobMSBShort(*image);  /* left_edge */
```
dlemstra, line 450: Wondering if we should set the `image->page`?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+  (void) ReadBlobMSBShort(*image);  /* top_edge */
+  img_width=ReadBlobMSBShort(*image);
```
dlemstra: Don't think we need the `img_` prefix?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+  img_height=ReadBlobMSBShort(*image);
+  img_depth=ReadBlobMSBShort(*image);
+  (void) ReadBlobMSBLong(*image);   /* image_data ptr */
+  plane_pick=(unsigned char) ReadBlobByte(*image);
+  plane_on_off=(unsigned char) ReadBlobByte(*image);
+  (void) ReadBlobMSBLong(*image);   /* next ptr */
```
dlemstra, line 458: Does this mean we only read the fist image?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+
+  ssize_t
+    y;
+
+  (void) ReadBlobMSBShort(*image);  /* left_edge */
+  (void) ReadBlobMSBShort(*image);  /* top_edge */
+  img_width=ReadBlobMSBShort(*image);
+  img_height=ReadBlobMSBShort(*image);
+  img_depth=ReadBlobMSBShort(*image);
+  (void) ReadBlobMSBLong(*image);   /* image_data ptr */
```
dlemstra, line 455: Can we ignore this?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+  (void) ReadBlobMSBLong(*image);   /* next ptr */
+  if (EOFBlob(*image) != MagickFalse)
+    return(MagickFalse);
+  if ((img_width == 0) || (img_height == 0) || (img_depth == 0) ||
+      (img_width > MaxAmigaIconDimension) ||
+      (img_height > MaxAmigaIconDimension) || (img_depth > 8))
+    return(MagickFalse);
+  words_per_row=((size_t) img_width+15)/16;
+  bytes_per_row=words_per_row*2;
+  if (HeapOverflowSanityCheckGetSize(bytes_per_row,(size_t) img_height,
```
dlemstra: Good to see you are adding these checks!

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+      sizeof(*planes[p_idx]));
+    if (planes[p_idx] == (unsigned char *) NULL)
+    {
+      size_t j;
+      for (j=0; j < p_idx; j++)
+        planes[j]=(unsigned char *) RelinquishMagickMemory(planes[j]);
+      planes=(unsigned char **) RelinquishMagickMemory(planes);
+      return(MagickFalse);
+    }
+    if (ReadBlob(*image,plane_size,planes[p_idx]) != (ssize_t) plane_size)
```
dlemstra: Could this not have been an `||` so we only need the  cleanup code once?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+          if (data_plane < (size_t) img_depth)
+            color_index|=(unsigned int)
+              (((planes[data_plane][byte_idx] >> bit_idx) & 1) << plane_num);
+          data_plane++;
+        }
+        else if (plane_on_off & (1 << plane_num))
+          color_index|=(1U << plane_num);
+      }
+      if (color_index == 0)
+        SetAmigaPixelRGBA(*image,q,0,0,0,0);
```
dlemstra: Maybe also rewrite this to make it more clear?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+        SetAmigaPixelRGBA(*image,q,0,0,0,255);
+      q+=(ptrdiff_t) GetPixelChannels(*image);
+    }
+    if (SyncAuthenticPixels(*image,exception) == MagickFalse)
+    {
+      render_status=MagickFalse;
+      break;
+    }
+  }
+  for (p_idx=0; p_idx < (size_t) img_depth; p_idx++)
```
dlemstra: Now starting to wonder if we should not have a separate cleanup method (not `inline`).

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+    *ni_palette,
+    *ni_pixels,
+    *line_bits;
+
+  if ((im_count == 0) || (im_lines == (char **) NULL))
+    return(MagickFalse);
+  first_line=im_lines[0];
+  if (strlen(first_line) < 5)
+    return(MagickFalse);
+  ni_transparent=((unsigned char) first_line[0] == 0x42) ? 1 : 0;
```
dlemstra: Can you do me a big favor and also replace the rest of the `0x42` notation to improve the readability.

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+    Read 78-byte DiskObject header (big-endian).
+  */
+  magic=ReadBlobMSBShort(image);
+  if (magic != AmigaIconMagic)
+    ThrowReaderException(CorruptImageError,"ImproperImageHeader");
+  (void) ReadBlobMSBShort(image);  /* version */
+  (void) ReadBlobMSBLong(image);   /* gadget next */
+  (void) ReadBlobMSBShort(image);  /* left_edge */
+  (void) ReadBlobMSBShort(image);  /* top_edge */
+  (void) ReadBlobMSBShort(image);  /* width */
```
dlemstra: Can we really ignore all these values? If we can that we can also do a seek instead?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+      **im1_lines = (char **) NULL,
+      **im2_lines = (char **) NULL;
+
+    newicon_failure=MagickFalse;
+    tooltypes_corrupt=MagickFalse;
+    tooltypes_resource_failure=MagickFalse;
+    if (EOFBlob(image) != MagickFalse)
+    {
+      tooltypes_corrupt=MagickTrue;
+      goto newicon_cleanup;
```
dlemstra: Not a fan of `goto` statements. Can you avoid that? Feels more like this part should have been a separate method?

coders/wbinfo.c
```diff
@@ -0,0 +1,1610 @@
...
+{
+  MagickOffsetType
+    offset;
+
+  MagickSizeType
+    blob_size;
+
+  offset=TellBlob(image);
+  blob_size=GetBlobSize(image);
+  if ((offset < 0) || ((MagickSizeType) offset > blob_size) ||
```
dlemstra, line 97: This makes no sense? And you also seem to be using this method in weird places.

dlemstra (commented):
Thanks for adding this for us. Left a bunch of comments. Feel free to take ask follow up questions in case I was not clear.

coders/wbinfo.c
bitplane: no this is junk raw memory pointer that gets re-written on icon load on the Amiga, we just check for select_render to see if there's a selected/clicked version

coders/wbinfo.c
bitplane: Yeah they're raw memory written to the file. The icons actually mutated on disk as you moved them around in AmigaOS (they act as the UI's proxy to otherwise hidden files in a dir, and were written at load time and re-read as a cache to save RAM), so there's unserialized data in there that we don't need - they're icon.library's pointers

coders/wbinfo.c
bitplane: Yep this one's more RAM junk - not part of the icon data, replaced with a seek here.

coders/wbinfo.c
bitplane: Ah yeah good point. Force of habit - no icon file is gonna be 4GB, the 32-bit size_t thing won't ever be a problem

coders/wbinfo.c
bitplane: I added it, but I think it's 0,0 in the vast majority of icons. I'll double check against the entire corpus though and see what they look like.

coders/wbinfo.c
bitplane: It drops a blank half-initialized frame when SetImageExtent fails.

coders/wbinfo.c
bitplane: Only 64 used these fields out of 297k icons, and they're all bad - in every case it pushes the image data out of bounds, and it differs between selected and ordinary icons.
It might have been used in weird layout hacks or with some other library override, but it looks like it'll be a source of corruption. I'll add a comment saying so.

coders/wbinfo.c
bitplane: Removed those, kept the useful ones :)

coders/wbinfo.c
bitplane: Oops, yes, nice catch. Fixed.

coders/wbinfo.c
bitplane: I think I got all of these!

coders/wbinfo.c
bitplane: Done. Arithmetic is decimals, but kept the declarations as hex

coders/wbinfo.c
bitplane: done, tidied it up and put it in the one place

coders/wbinfo.c
bitplane: Makes sense, added a RelinquishPlanes() shared by all three cleanups (alloc/read failure, append failure, end-of-render free)

coders/wbinfo.c
bitplane: Moved to ReadToolTypesNewIcons, so no more gotos. The FORM/ICON chunk parsing got the same treatment

coders/wbinfo.c
bitplane: Made it more explicit: palette/fallback branch for RGB, separate branch for alpha

bitplane:
Thanks for the review! I think I addressed all your comments, please re-open if I missed anything :)

I did a full conversion sweep against the full 300k and looked over a few thousand, they  look unchanged, and re-ran the main test data from the Python loader:

https://github.com/user-attachments/assets/9ed7769c-f940-47f7-96dc-86cdbb9b3d33

I did find a couple of hundred that were missing trailing metadata, ones that an Amiga would have loaded but I was failing on, so I've added a warning for those and let it just load the available data as frames.

dlemstra (approved):
Still some things left but we can do that after the merge

PR #8628, opened 2026-03-19
Fix NULL pointer dereference in ReadDOTImage()

hirorogo:
## Summary

`GetBlobFileHandle()` can return NULL when the image blob is opened in a non-file mode (e.g., when gzip/bzip2 magic bytes are detected by `OpenBlob()`). In `ReadDOTImage()`, the return value was passed directly to `agread()` without a NULL check, causing a NULL pointer dereference crash.

This adds a NULL guard before calling `agread()`, returning early with proper cleanup consistent with the existing error handling pattern in the function.

## Test plan

- [ ] Open a `.dot` file whose first bytes are gzip magic (`0x1F 0x8B 0x08`) — should return gracefully instead of crashing
- [ ] Normal `.dot` file processing still works


dlemstra:
This coder is registered with `entry->flags^=CoderBlobSupportFlag;` and that makes sure there will always be a file. So we don't need this change.

Issue #8104, opened 2025-04-15
Orientation is not preserved when converting from ARW to JXL


uhthomas:
### ImageMagick version

7.1.1-47

### Operating system

Linux

### Operating system, version and so on

Arch Linux, latest

### Description

Orientation metadata is not preserved when converting from Sony Raw (ARW) to JPEG XL (JXL).

```sh
Version: ImageMagick 7.1.1-47 Q16-HDRI x86_64 22763 https://imagemagick.org
Copyright: (C) 1999 ImageMagick Studio LLC
License: https://imagemagick.org/script/license.php
Features: Cipher DPC HDRI Modules OpenCL OpenMP(4.5) 
Delegates (built-in): bzlib cairo djvu fftw fontconfig freetype heic jbig jng jp2 jpeg jxl lcms lqr ltdl lzma openexr pangocairo png raqm raw rsvg tiff uhdr webp wmf x xml zip zlib zstd
Compiler: gcc (14.2)
```

### Steps to Reproduce

Convert an ARW image with orientation metadata to JXL and observe that the rotation metadata is not preserved.

```sh
❯ magick mogrify -format jxl _DSC8360.ARW 
```

### Images

I'm sorry, the images my camera produces are 130MB and will not fit in the file size limits.

uhthomas:
I hope this helps.

```sh
❯ magick identify -verbose _DSC6872.ARW | rg -i orientation
  Orientation: LeftBottom
❯ magick identify -verbose _DSC6872.jxl | rg -i orientation
  Orientation: TopLeft
```

uhthomas:
It seems the behavior isn't consistent across different target formats?

```sh
❯ magick identify -verbose _DSC6872.png | rg -i orientation
  Orientation: LeftBottom
❯ magick identify -verbose _DSC6872.jpg | rg -i orientation
  Orientation: Undefined
```

uhthomas:
Ah, I need `-auto-orient` - not sure why this isn't the default :(

Issue #8264, opened 2025-07-20
Images converted to JXL are not reoriented


uhthomas:
### ImageMagick version

 7.1.1-47,

### Operating system

Linux

### Operating system, version and so on

Arch Linux

### Description

Modern formats like JXL and HEIF completely ignore exif data for orientation and instead rely on the image being oriented correctly.

https://github.com/libjxl/libjxl/blob/b7961e1318634a95c7d689ec340dbad8a42d4ed4/lib/include/jxl/decode.h#L1251-L1254

Images converted with imagemagick, JPG to JXL are not automatically reoriented, which sort of violates the spec. I understand that there is an `auto-orient` flag from https://github.com/ImageMagick/ImageMagick/issues/8104, though I believe this should be default for JXL and HEIF given that spec compliant decoders will ignore exif metadata. Another point of reference is that `cjxl` will correctly reorient the original image.

### Steps to Reproduce

Convert an image from JPG to JXL and observe that orientation is not preserved, because the exif data is ignored by compliant decoders.

### Images

My images are too big, sorry.

fmw42:
Imagemagick does not automatically rotate any format images.  You have to explicitly use -auto-orient in your command.

`magick image.jpg -auto-orient image.jxl`

Does that not work?  How big are your images?  Can you crop one and post the smaller crop?

uhthomas:
I understand, though I think there needs to be some exceptions for formats which explicitly ignore exif data for orientation. cjxl should be a reference for how images should be converted, and imagemagick does not align with that. I am not sure I can think of any scenarios where it would be intentional to omit `auto-orient` when converting to jxl or heif. 

My images are about 40MBs, I can crop and demonstrate if that would be helpful?

fmw42:
Why would Imagemagick ignore the exif data for orientation in just jxl?  Do you think there is a bug?

Yes a demonstration on a subset would be helpful or just upload your 40 Mb image to some free hosting service and put the URL here

uhthomas:
So I have a reference image, `DSC06595-scaled.JPG`, one of which I have converted with imagemagick, and the other with cjxl.

```sh
❯ magick mogrify -format jxl DSC06595-scaled.JPG 
```

```sh
❯ cjxl DSC06595-scaled.JPG DSC06595-scaled-cjxl.jxl
```

It is clear that the original image, and the image converted with imagemagick both have an orientation of 8 (90 degrees counter-clockwise).

```sh
[...]

PR #8673, opened 2026-04-15
Fix JXL orientation when writing

alkino:
### Prerequisites

### Description
JXL documentation told us to ignore EXIF because orientation is encoded in the stream itself.
ImageMagick was keeping the EXIF but not setting the stream. Both should be kept and synchronized.


alkino:
Should fix #8104 and #8264

dlemstra (approved):
Thanks for adding this for us!

PR #8703, opened 2026-04-28
Add -strict option to treat coder warnings as errors

### LINKED ISSUE #8682: magick should throw error if compression method is not supported
### ImageMagick version

7.1.2-17

### Operating system

Linux

not relevant

### Description

`magick` silently fails
when a compression method is not supported

related #8386 

### actual

output image is written without compression

```console
$ magick rose: -compress JBIG2 test.jbig2.tiff

$ echo $?
0

$ tiffinfo test.jbig2.tiff | grep Compression
  Compression Scheme: None
```

### expected

`magick` should fail

```console
$ magick rose: -compress JBIG2 test.jbig2.tiff
error: JBIG2 compression is not supported in TIFF format

$ echo $?
1
```

### Images

`rose:`

eyupcanakman:
### Prerequisites

### Description

Reworked per @dlemstra's review (#8703 (comment)).

`-strict` is now a global flag that turns coder warnings into errors instead of a TIFF-only check. The TIFF writer throws `CoderWarning` when a requested compression is not configured, and `ThrowMagickExceptionList` bumps the severity to the matching error band whenever `SetWarningsAsErrors(MagickTrue)` is set. Other coders gain the same behavior under `-strict` once they adopt the `if (exception->severity >= ErrorException)` cleanup-and-return pattern already used in `svg.c`, `wpg.c`, and `msl.c`.

Closes #8682


dlemstra:
This option now only works for the tiff coder. This should then be a define for that specific coder instead. I am wondering if we should choose a different approach. Maybe we should print a warning here and use your `-strict` idea to treat warnings as errors?

eyupcanakman:
Agree, warnings-as-errors makes sense. I'll rework it so the tiff coder emits a warning and `-strict` escalates it.

eyupcanakman:
Pushed the rework. `-strict` is now a global flag and the warning-to-error promotion happens inside `ThrowMagickExceptionList`. The TIFF writer emits `CoderWarning` for an unconfigured codec and bails out when the severity has been promoted.

eyupcanakman:
Pushed the warnings-as-errors version in May, the way you suggested. Can change the approach if you'd rather.

Issue #8728, opened 2026-05-13
build error with libheif 1.20


remicollet:
### ImageMagick version

7.1.2-22

### Operating system

Linux

### Operating system, version and so on

Fedora 43

### Description

```
coders/heic.c: In function 'ReadHEICSequenceFrames':
coders/heic.c:699:17: error: 'struct heif_decoding_options' has no member named 'ignore_sequence_editlist'
  699 |   decode_options->ignore_sequence_editlist=1;
      |                 ^~

```

### Steps to Reproduce

Reproduced only with 1.20

Notice: 
- build OK with libheif < 1.20 (Tested with 1.17, 1.147...)
- build OK with libheif 1.21

### Images

N/A

PR #8729, opened 2026-05-13
ignore_sequence_editlist only in libheif 1.21

remicollet:
See #8728


remicollet:
See https://github.com/strukturag/libheif/blob/master/libheif/api/libheif/heif_decoding.h#L112

`ignore_sequence_editlist` is defined in `version 8 options (v1.21.0)`

PR #8750, opened 2026-05-22
reject farbfeld files with zero columns or rows

dxbjavid:
ReadFARBFELDImage reads width and height directly from the blob into image->columns and image->rows and then takes the image_info->ping early-return path before SetImageExtent gets a chance to reject zero dimensions, so a 16-byte file with width=0 or height=0 (or a truncated header where ReadBlobLong returns 0) succeeds and `identify` reports a 0x0 image. Same shape as the DCM fix in 84fbcef referencing GHSA-8pj9-6897-74xc.


dlemstra (approved):
Not really an issue because of the ping but we could add a check here as you are suggesting. Thanks!

PR #8760, opened 2026-05-26
reject fits files with an invalid bits per pixel

jmestwa-coder:
ReadFITSImage breaks out of the header loop for a non-extension file before the bits_per_pixel and dimension checks, so a simple FITS declaring BITPIX 7 reads as a 7-bit image rather than being rejected. Move the extend break below those checks so the primary header gets validated like the extensions already are.

### DIFF
--- a/coders/fits.c
+++ b/coders/fits.c
@@ -226,7 +226,6 @@ static Image *ReadFITSImage(const ImageInfo *image_info,
   typedef struct _FITSInfo
   {
     MagickBooleanType
-      extend,
       simple;
 
     int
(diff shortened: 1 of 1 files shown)


dlemstra (commented):
I am not sure why we had that check but it makes no sense to me. Can you update your pull request to completely remove the `extend` field from the `FITSInfo`?

jmestwa-coder:
Done. Pulled the extend field out of FITSInfo; the loop now stops on the existing simple/naxis check, which already covers the primary header. Valid 8/16/32-bit and float FITS still read and the BITPIX=7 case is rejected.

dlemstra (approved):

PR #8763, opened 2026-05-26
reject tim2 files with zero columns or rows

dxbjavid:
ReadTIM2Image sets image->columns and image->rows from the header but only calls SetImageExtent (which rejects zero) inside ReadTIM2ImageData, which the ping path skips, so `identify -ping` on a TIM2 declaring zero width or height reports a 0x4 image instead of failing. Add the zero check to the ping branch, matching 6eb7297.

### DIFF
--- a/coders/tim2.c
+++ b/coders/tim2.c
@@ -762,6 +762,8 @@ static Image *ReadTIM2Image(const ImageInfo *image_info,
         if (status==MagickFalse)
           break;
       }
+    if ((image->columns == 0) || (image->rows == 0))
+      ThrowReaderException(CorruptImageError,"ImproperImageHeader");
     if ((image_info->ping != MagickFalse) && (image_info->number_scenes != 0))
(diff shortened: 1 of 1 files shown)


coders/tim2.c
```diff
@@ -762,6 +762,9 @@ static Image *ReadTIM2Image(const ImageInfo *image_info,
         if (status==MagickFalse)
           break;
       }
+    else
```
dlemstra: Please move this lower and do it in the before the `break`.

dxbjavid: Moved it down to just before the break. I kept it unconditional rather than nesting it in the number_scenes != 0 check though: tim2 doesn't hit SetImageExtent on the ping path the way sgi does, so a plain `identify -ping` with no scene range would still come back as 0x4. This way it's caught either way.

dlemstra:

PR #8765, opened 2026-05-27
reject hdr files with a negative width or height

dxbjavid:
Noticed ReadHDRImage parses the resolution line width and height as signed int but only the zero case is rejected later. A negative value casts to a huge size_t and slips past that check, so a `-Y -1 +X 4` header makes identify -ping surface a 4x18446744073709551616 image. Reject non-positive dimensions at the parse site, like xbm.c already does.

### DIFF
--- a/coders/hdr.c
+++ b/coders/hdr.c
@@ -340,6 +340,9 @@ static Image *ReadHDRImage(const ImageInfo *image_info,ExceptionInfo *exception)
 
                   if (MagickSscanf(value,"%d +X %d",&height,&width) == 2)
                     {
+                      if ((width <= 0) || (height <= 0))
+                        ThrowReaderException(CorruptImageError,
+                          "ImproperImageHeader");
                       image->columns=(size_t) width;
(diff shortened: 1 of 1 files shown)


coders/hdr.c
dlemstra, line 363: There are already checks here and below?

dxbjavid: Those checks look like they are  operating on image->columns and image->rows after assignment. My concern is specifically about negative values during parsing: the dimensions are read as signed integers, but once a negative value is assigned to the image fields it can become a very large positive value due to the signed to unsigned conversion.

So a header like -Y -1 +X 4 can bypass the later zero checks because the value is no longer negative at that point. The added validation rejects non-positive dimensions before the conversion happens and keeps the behavior consistent with other coders such as xbm.c.

dlemstra: Fair, it might be a good idea to disallow that.

dlemstra (approved):

PR #8767, opened 2026-05-27
reject icon files with a negative width or height

dxbjavid:
ReadICONImage reads the bitmap width and height as signed longs at icon.c:392 but never rejects negatives, so a file with a negative biWidth makes `identify 'ICON:file[0]'` surface an 18446744073709551616-wide image because the ping-with-scene path returns before SetImageExtent rejects it. Reject negative width and height in the bitmap branch before they are cast to size_t.

### DIFF
--- a/coders/icon.c
+++ b/coders/icon.c
@@ -457,6 +457,8 @@ static Image *ReadICONImage(const ImageInfo *image_info,
 
         if (bits_per_pixel > 32)
           ThrowICONReaderException(CorruptImageError,"ImproperImageHeader");
+        if ((width < 0) || (height < 0))
+          ThrowICONReaderException(CorruptImageError,"ImproperImageHeader");
         (void) ReadBlobLSBLong(image); /* compression */
         (void) ReadBlobLSBLong(image); /* image_size */
         (void) ReadBlobLSBLong(image); /* x_pixels */


coders/icon.c
```diff
@@ -457,6 +457,8 @@ static Image *ReadICONImage(const ImageInfo *image_info,
 
         if (bits_per_pixel > 32)
           ThrowICONReaderException(CorruptImageError,"ImproperImageHeader");
+        if ((width < 0) || (height < 0))
```
dlemstra, line 460: This is incorrect? See line 472 and 474? And I don't think we need these checks anymore?

dxbjavid: Good catch. I may have overlooked the logic around lines 472 and 474 when adding this check.

My concern was that a negative width could be converted to a large 'size_t' value before the identify path reaches 'SetImageExtent()', but I'll recheck the current flow to verify whether that's still possible. If the existing logic already prevents that, I'll drop this change.

dxbjavid: You're right on both counts. Walked through it again: on the actual read path the dimension goes through GetICONSize and SetImageExtent already rejects it, so the only thing left was the huge number in the identify ping output, which is just the basic header sanity check rather than a real decode problem. And rejecting a negative height is wrong anyway since that's the top-down marker in the BMP encoding. Dropping the check and closing this.

PR #8768, opened 2026-05-28
reject tim files with zero columns or rows

dxbjavid:
Hit this fuzzing `identify 'TIM:file[0]'` on a no-CLUT TIM. ReadTIMImage reads the actual image columns and rows after the CLUT block, but the (image_info->ping && number_scenes) early break exits the do/while before that read, so on a no-CLUT TIM image->columns and image->rows are still zero and a 0x0 image is reported instead of failing. Adds the zero check to the early break, matching the pict.c part of 6eb7297.


dlemstra:

PR #8769, opened 2026-05-28
reject pict clip rectangles with zero width or height

dxbjavid:
Found this poking at PICT clipping opcodes with identify -ping. ReadRectangle rejects right<left and bottom<top but not the equal case, so the opcode 0x01 branch in pict.c assigns 0 to image->columns/rows via the frame subtraction when right==left or bottom==top. SetImageExtent on the full-read path catches it at pict.c:1025, but the image_info->ping break at pict.c:1023 returns first, so identify reports the file as 'PICT 0x0' (confirmed: 'PICT 0x0' exit 0 before, ImproperImageHeader after; PerlMagick/t/input.pict still pings and decodes at 70x46). Same idea as the main-header guard added in 6eb7297.


dlemstra:

PR #8771, opened 2026-05-28
reject uhdr files with zero columns or rows

dxbjavid:
ReadUHDRImage in coders/uhdr.c assigns image->columns/rows from uhdr_dec_get_image_width/height and then takes the image_info->ping early return before SetImageExtent runs. If the decoder reports a zero dimension, the ping path surfaces a 0xN image to callers while the full read path is still rejected later by SetImageExtent. Adds the zero-dimension check inside the ping return block, mirroring the cin/dpx/farbfeld/pdb/pict/rgf/rla hunks of commit 6eb7297.

### DIFF
--- a/coders/uhdr.c
+++ b/coders/uhdr.c
@@ -202,6 +202,11 @@ static Image *ReadUHDRImage(const ImageInfo *image_info,
 
   if (image_info->ping != MagickFalse)
   {
+    if ((image->columns == 0) || (image->rows == 0))
+    {
+      uhdr_release_decoder(handle);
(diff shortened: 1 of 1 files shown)


urban-warrior:
identify reliably reports the image filename, width, height, and depth. It only performs basic sanity checks. For full checks, use

```
$ identify +ping image.udhr
```

It will then fully validate the image and report an exception. This is expected behavior.

netliomax25-code:
Makes sense, the ping path is meant to stay lightweight and +ping does the real validation. I'll close this.

PR #8776, opened 2026-05-29
reject empty string when pinging a strimg image

dxbjavid:
ReadSTRIMGImage sets image->columns from strlen of the interpreted string and rows to one, then takes the image_info->ping early return before SetImageExtent runs. An empty STRIMG: string leaves columns at zero, so a full read is rejected by SetImageExtent with NegativeOrZeroImageSize while identify -ping happily reports a 0x1 image. Reject the zero-width case in the ping branch so both paths agree, matching the guards added in 6eb7297.


dlemstra:

PR #8777, opened 2026-06-01
fix out-of-bounds read in polynomial distort with no arguments

dxbjavid:
GenerateCoefficients reads arguments[0] at distort.c:426 for the Polynomial method before checking number_arguments, so DistortImage with zero arguments (`-distort Polynomial ''`) reads out of bounds. Add the count check before the read, matching the cylinder/plane fix in GHSA-p9rq-q46c-g4x6.


dlemstra:
Another pull request has already been merged to fix this.

PR #8781, opened 2026-06-02
reject cineon files with an invalid bits per pixel

dxbjavid:
cin.c:725 sets image->depth from the channel bits_per_pixel byte without checking it, so a Cineon file declaring 0 or a value above 32 is accepted and only fails later (allocation failure, or a truncated read mid-decode). Add the same depth==0 || depth>32 rejection that the sibling dpx reader already does.


dlemstra:
We now have a case that allows 64-bit depth. Should we remove that or is that valid and should your check be changed?

dxbjavid:
Good point, the 64-bit case in GetBytesPerRow is a legitimately handled depth so rejecting it would be wrong. I've raised the upper bound to 64 to match, so 0 and anything above 64 are still rejected at parse time but a valid 64-bit file is left alone.

dlemstra (approved):

dlemstra (approved):

dxbjavid:
thanks for the review and merge.

dlemstra:
Thanks for helping us!

PR #8782, opened 2026-06-02
fix oob read of gps rationals in GetEXIFProperty

dxbjavid:
GetEXIFProperty reads three rationals (24 bytes) for the GPS latitude, longitude and timestamp tags but only validates the entry's declared component count first. A GPS tag declaring fewer than three rationals reads past the validated EXIF directory extent. Require the 24 bytes before reading, matching the other range checks here.


dxbjavid:
Thanks for the review and merge! Appreciated

PR #8786, opened 2026-06-04
feat(8785) Downgrade ubuntu container to 18.10

### LINKED ISSUE #8785: Support older versions of libc for precompiled AppImage file
### Is your feature request related to a problem? Please describe.

First off, the AppImage is really helpful for getting everything running on a system without needing to worry about incompatible dependencies, thanks for adding that.

I have some older AlmaLinux 8 servers which I want to run the AppImage on, however the precompiled image is using Ubuntu 22, which comes with glibc 2.35.
Any RHEL server would need the latest v10 upgrade just to get a supported glibc.

It would be helpful to target a lower version of glibc.
Obviously the ideal scenario would be to upgrade the servers instead, but it would be more useful broadly to have a lower dependency version, so it can be used without needing the latest OS upgrades.

### Describe the solution you'd like

Downgrade the container image which builds the appImage to Ubuntu 18.10 (which comes with glibc 2.27)

### Describe alternatives you've considered

Upgrading the server instead, which comes with risks of breaking changes

### Additional context

_No response_

caymaynard:
### Prerequisites

### Description
Downgrades the container version for building precompiled binaries for better compatibility on older servers.

closes #8785

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -440,6 +440,12 @@ jobs:
             packages: clang
 
     steps:
+
+      - name: fix package repo
+        run: |
+          sed -i 's|http://.*archive.ubuntu.com|http://old-releases.ubuntu.com|g' /etc/apt/sources.list
+          sed -i 's|http://security.ubuntu.com|http://old-releases.ubuntu.com|g' /etc/apt/sources.list
+
       - name: Install dependencies
         run: |
           set -e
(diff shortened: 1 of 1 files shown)


dlemstra:
Thanks for opening a PR But this is but this is probably not how we would want to support this if we would even want to do that. If we would do this it would be a separate build. We will keep your issue open but close this PR until we know what we would want to do with this.

Issue #8744, opened 2026-05-19
Batch processing of gain maps does not appear to be possible with IM


gregbenz:
### ImageMagick version

7.1.2-23

### Operating system

MacOS

### Operating system, version and so on

26.5 

### Description

Batch conversion with normal ImageMagick commands like mogrify -resize 50% *.jpg only uses the standard JPEG coder. 

It sees and resizes the SDR/base JPEG, but it does not preserve the gain map JPEG aux image or metadata. The result is an SDR JPEG, not a gain-map JPEG (the HDR content is stripped).

### Steps to Reproduce

Compile with gain map support:
```
PKG_CONFIG=/private/tmp/im-release-testing/tools/pkg-config \
IM_RELEASE_LOCAL=/private/tmp/im-release-testing/local \
CPPFLAGS="-I/opt/homebrew/opt/jpeg-turbo/include" \
LDFLAGS="-L/opt/homebrew/opt/jpeg-turbo/lib -L/private/tmp/im-release-testing/local/lib -Wl,-rpath,/private/tmp/im-release-testing/local/lib" \
/private/tmp/im-release-testing/src/ImageMagick-7.1.2-23/configure \
  --with-uhdr=yes \
  --prefix=/private/tmp/im-release-testing/install
```

Transcode a JPG gain map using mogrify, the output will no longer have a gain map.

### Images

Any gain map could be used, nothing special about this source:
<img width="1080" height="1080" alt="Image" src="https://github.com/user-attachments/assets/7ae472ef-319d-4132-ac13-9a7506b2e973" />

urban-warrior:
Download the latest beta release (building now) and let us know if the issue persists.

gregbenz:
Analyzing the official 7.1.2-26 beta source: 

The intended fix is visible:

1. Read path: ReadUHDRImage() extracts the compressed gain map and metadata from libuhdr, then stores them on the IM image as an hdrgm profile plus hdrgm:* properties. See uhdr.c (line 228).
2. Write path: WriteUHDRImage() looks for that hdrgm profile and tries to pass it back to libuhdr instead of regenerating the gain map. See uhdr.c (line 668).

Potential reasons mogrify does not show gain map output:

1. The fix is inside the UHDR coder, but normal mogrify *.jpg **still routes files as JPEG, not UHDR**. 
UHDR is registered as its own format at uhdr.c (line 438). 
[...]

Issue #8743, opened 2026-05-19
IM corrupts gain maps when transcoding


gregbenz:
### ImageMagick version

7.1.2-23

### Operating system

MacOS

### Operating system, version and so on

26.5

### Description

When transcoding an existing JPG gain map, the output is significantly degraded. The following issues are observed:

- The gain map metadata appears to be completely regenerated / altered, rather than preserved.
- The gain map min / max gets expanded from 0-6.8 to -14.3 - +7.9. This stretches the same bits over 3x and causes significant banding.
- The gain map shows significant artifact (notice the dark halos around the stars and black spots in the gain map itself)
- The HDR headroom resets from actual (4 stops here) to 5.6 stops). This causes the image to darken significantly.

The gain map metadata (headroom, etc) should not be altered by simple transcoding (including resampling or cropping).

Transcoding with other libraries using libultrahdr (Sharp and wasm-vips) shows much higher quality output. Please use the P3 sample at https://gregbenzphotography.com/test/wasm-vips-hdr-demo/ with 50% scale percent and 90% quality to compare to the command below.

https://gregbenzphotography.com/hdr-soft-proofing-test-hdr-gain-maps/ may be used to directly compare the two output images.

### Steps to Reproduce

Compile with gain map support:
```
PKG_CONFIG=/private/tmp/im-release-testing/tools/pkg-config \
IM_RELEASE_LOCAL=/private/tmp/im-release-testing/local \
CPPFLAGS="-I/opt/homebrew/opt/jpeg-turbo/include" \
LDFLAGS="-L/opt/homebrew/opt/jpeg-turbo/lib -L/private/tmp/im-release-testing/local/lib -Wl,-rpath,/private/tmp/im-release-testing/local/lib" \
/private/tmp/im-release-testing/src/ImageMagick-7.1.2-23/configure \
  --with-uhdr=yes \
  --prefix=/private/tmp/im-release-testing/install
```

Transcode per:
```
./magick \
   \( -define uhdr:output-color-transfer=srgb "UHDR:./source/P3.jpg" -resize 50% -depth 8 \) \
  \( -define uhdr:output-color-transfer=linear "UHDR:./source/P3.jpg" -resize 50% -depth 16 \) \
  -define uhdr:hdr-color-transfer=linear \
  -quality 90 \
  -define uhdr:gainmap-quality=90 \
  "UHDR:./output/P3-50pct.jpg"
```

### Images

[...]

PR #8787, opened 2026-06-07
Preserve Ultra HDR JPEG gain maps when transcoding

gregbenz:
### Prerequisites

### Description

This updates Ultra HDR / ISO gain-map JPEG handling so existing gain-map JPEGs can be transcoded through normal JPEG commands while preserving the gain map.

Changes include:

- Detect Ultra HDR JPEGs during normal JPEG reads and route those files through the UHDR coder.
- Route JPEG writes through the UHDR coder when the image contains a preserved `hdrgm` profile.
- Preserve compressed gain-map data and metadata when reading Ultra HDR JPEGs.
- Preserve gain maps for supported geometry operations: resize, crop, flip/flop, and 90-degree rotations.
- Fail safely for unsupported geometry operations instead of writing a stale or mismatched gain map.
- Use libultrahdr’s compressed base-image plus compressed gain-map path instead of regenerating a gain map from a reconstructed HDR representation.
- Preserve ICC profiles from Ultra HDR JPEGs, including Display P3 inputs. The libultrahdr ICC APP2 wrapper is stripped before storing the raw ICC profile in ImageMagick.

Validation performed with P3 and sRGB ISO gain-map JPEGs:

- `magick input.jpg -resize 50% output.jpg` preserves Ultra HDR output.
- `magick mogrify -path out -resize 50% *.jpg` preserves Ultra HDR output.
- Outputs retain MPF with 2 images.
- Embedded gain maps are resized from `1080x1080` to `540x540` for a 50% resize.
- P3 output ICC profile matches the source ICC profile byte-for-byte.
- Visible SDR base output matches an ordinary JPEG resize.

Related issues: #8743, #8744

### DIFF
--- a/MagickCore/profile-private.h
+++ b/MagickCore/profile-private.h
@@ -33,6 +33,8 @@ extern MagickExport StringInfo
     ExceptionInfo *exception);
 
 extern MagickPrivate void
+  AppendImageProfileProperty(Image *,const char *,const char *,const char *,
+    ExceptionInfo *),
   Update8BIMClipPath(const Image *,const size_t,const size_t,
     const RectangleInfo *),
   SyncImageProfiles(Image *);
--- a/MagickCore/profile.c
+++ b/MagickCore/profile.c
@@ -92,6 +92,67 @@ static void
   WriteTo8BimProfile(Image *,const char*,const StringInfo *);
 
 /*
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                             %
+%                                                                             %
+%                                                                             %
+%   A p p e n d I m a g e P r o f i l e P r o p e r t y                       %
+%                                                                             %
+%                                                                             %
+%                                                                             %
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%
+%  AppendImageProfileProperty() appends a semicolon-delimited value to an image
+%  property when the image contains the requested profile.
+%
+%  The format of the AppendImageProfileProperty method is:
+%
+%      void AppendImageProfileProperty(Image *image,const char *profile,
+%        const char *property,const char *value,ExceptionInfo *exception)
+%
+%  A description of each parameter follows:
+%
+%    o image: the image.
+%
+%    o profile: the image profile name.
+%
+%    o property: the image property name.
+%
+%    o value: the property value to append.
+%
+%    o exception: return any errors or warnings in this structure.
+%
+*/
+MagickPrivate void AppendImageProfileProperty(Image *image,const char *profile,
+  const char *property,const char *value,ExceptionInfo *exception)
+{
+  char
+    *property_value;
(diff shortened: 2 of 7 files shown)


parasol-aser:
**`coders/jpeg.c` — `IsUltraHDRJPEGImage` / `ReadJPEGImage` dispatch (~L1810, sniff at ~L272):** every JPEG read now calls `is_uhdr_image` on the raw blob and re-dispatches matching images into libultrahdr's full ISO-BMFF/MPF/XMP decoder, with no runtime opt-out — it's gated only by the compile-time `MAGICKCORE_UHDR_DELEGATE`. Previously libultrahdr was reachable only via an explicit `UHDR:` request, so this moves a large third-party C++ parser into the default `.jpg` attack surface: any present or future memory-safety bug there becomes reachable by ordinary JPEG decoding, plus every decode now pays an extra full-blob sniff. The heavy decode still passes through `ReadImage`/coder-policy checks and the in-repo guards (`data != NULL`, `length <= INT_MAX`) are correct, but a `policy.xml` rule denying the `UHDR` coder won't stop the pre-dispatch sniff, and operators can't keep plain libjpeg decoding without rebuilding.

Fix: add a runtime opt-out (e.g. a `-define jpeg:detect-uhdr=false` / coder hint, and/or honor coder-rights for `UHDR` inside `IsUltraHDRJPEGImage` before calling `is_uhdr_image`), and keep libultrahdr pinned/patched now that it's in the default path.

urban-warrior:
@gregbenz, could you split this pull request into two separate requests:

- one for updates to the UHDR coder
- one for detection of UHDR in the JPEG coder

We plan to apply the UHDR coder updates immediately. However, the UHDR detection in the JPEG coder has broader implications, as @parasol-aser noted, so we will need to approach that feature more carefully.

gregbenz:
@urban-warrior @parasol-aser  Thank you for the quick review. No problem, I have broken out "phase 1" changes in this PR: https://github.com/ImageMagick/ImageMagick/pull/8788.

I presume it would be best to wait to open a PR for "phase 2" (detection of UHDR in the JPEG coder) until after phase 1 is reviewed / merged to main, but please let me know if your preference is to approach that PR differently.

@parasol-aser Great feedback. I will revise per your input for "phase 2" prior to opening that PR.

urban-warrior:
Let’s avoid encumbering the JPEG coder with a dependency on libultra. 

A more robust approach is to detect UltraHDR in the JPEG coder during `ReadAPPProfiles()` by scanning for the HDR gain‑map namespaces:

- `http://ns.adobe.com/hdr-gain-map/1.0/`
- `http://ns.google.com/photos/1.0/gainmap/`

If either profile is present, set the `jpeg:hdrgm` artifact to indicate that the file contains UltraHDR metadata. You may also check for the Google UHDR APP11 segment as an additional signal.

After `jpeg_read_header()` completes, test for the `jpeg:hdrgm` artifact.   If it is set, redirect the image to the UHDR coder instead of the standard JPEG pipeline. This keeps the JPEG decoder lightweight while still enabling automatic UltraHDR detection based on embedded metadata rather than filename conventions or external hints.

gregbenz:
@urban-warrior Thank you for the review and approving the first phase of changes. I'm reviewing all the input here for revisions for the next phase and appreciate the clear guidance.

gregbenz:
closing and will submit a clean PR for phase 2

PR #8808, opened 2026-06-19
OSS-Fuzz: Fix build errors

arthurscchan:
This PR fixes the OSS-Fuzz fuzzer build on disabling the CLI tools for libtiff build that is currently failing and does not affect fuzzing.


dlemstra (approved):
It looks like this was already resolved but it might be a good idea to perform the build without this.

PR #8812, opened 2026-06-21
check rows*columns overflow in sct ReadSCTImage

dxbjavid:
### Prerequisites

- [x] I have verified that there are no overlapping pull-requests open

### Description

ReadSCTImage works out the expected pixel size from image->rows*image->columns*separations, but the rows and columns are taken straight from the textual SCT header fields, so on a crafted file the size_t multiply can wrap and slip past the extent > GetBlobSize check, after which SetImageExtent is reached with the oversized dimensions. This is the same overflow that was recently addressed in the vicar reader. I have guarded the multiplication with HeapOverflowSanityCheckGetSize and moved the size check ahead of SetImageExtent so an over-large file is rejected before any pixel cache is allocated.

### DIFF
--- a/coders/sct.c
+++ b/coders/sct.c
@@ -145,6 +145,7 @@ static Image *ReadSCTImage(const ImageInfo *image_info,ExceptionInfo *exception)
     *q;
 
   size_t
+    number_pixels,
     separations,
     separations_mask,
     units;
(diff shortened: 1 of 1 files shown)


dxbjavid:
Thanks @urban-warrior.

Issue #7715, opened 2024-10-22
AVIF images are corrupted (NCLX / CICP values improperly change for HDR or wide gamut images)


gregbenz:
### ImageMagick version

7.1.1-39 Q16-HDRI aarch64 22428

### Operating system

MacOS

### Operating system, version and so on

15.0.1 (24A348)

### Description

When an AVIF image is processed, the NCLX (CICP) values are being changed. This causes a wide gamut (P3) image to be interpreted as sRGB, resulting in the image being desaturated (looks washed out).

Unless there is an intentional change in color profile, these values should never change. The default behavior should be that the NCLX values are the same.

### Steps to Reproduce

The attached image was processed using:
`magick big.avif -resize 50% small.avif
`
You can then test for proper conversion using EXIFTOOL. The output should match the source for all reported values when testing:
`exiftool -colorprofiles -colorprimaries -transfercharacteristics -matrixcoefficients -videofullrangeflag myFile.JPG`

In the attached example, the source color primaries are changing from "SMPTE EG 432-1" (ie P3) to "BT.709" (ie the sRGB primaries). 

For the HDR example, the transfer characteristic is changing from "SMPTE ST 2084, ITU BT.2100 PQ" (ie PQ) to "sRGB or sYCC". This this is causing HDR content to be compressed to SDR, which creates a larger error.

### Images

[badAvifConversionFromBigToSmall.zip](https://github.com/user-attachments/files/17480417/badAvifConversionFromBigToSmall.zip)

gregbenz:
If there are function calls which may be used with IM to force preservation of the CICP values, it would be helpful to know as AVIF uploads to WordPress show this corruption when using IM to process the images: https://core.trac.wordpress.org/ticket/60889#comment:10 (it would help to know the command line equivalent of such function calls to due some testing directly with IM).

Such a workaround from WP may be helpful to offer more immediate mitigation, as the time for any IM updates to propagate across WP hosting environments is likely significant.

gregbenz:
@ farindk, any suggestions on option to decode / encode with libheif in a way which would preserve NCLX values when processing an AVIF?

farindk:
This problem has been addressed in the upcoming v1.19.0. It is already corrected in the master branch.
Previously, decoding often went through a forced RGB conversion internally in libheif, which changed the colorspace to sRGB.

Now, the original nclx values are preserved.
[...]

PR #8824, opened 2026-06-25
Preserve CICP and CLLI metadata

gregbenz:
## Summary

This preserves HEIC/AVIF color and HDR metadata when an image is read and then written again.

Specifically, this adds support for reading and writing CICP/NCLX color information, including color primaries, transfer characteristics, matrix coefficients, and full-range flag. It also adds support for CLLI content light level metadata, including MaxCLL and MaxFALL.

This helps address metadata loss which causes loss of wide gamut color and HDR as reported in ImageMagick/ImageMagick#7715.

## Root Cause

The HEIC reader imported raw ICC profiles, EXIF, and XMP, but did not import NCLX/CICP or CLLI item properties into the ImageMagick image model.

The HEIC writer already supported an explicit `heic:cicp` define, but it only wrote CICP when provided manually. There was no corresponding source-property preservation path, and no CLLI write support.

As a result, transcoding or resizing HEIC/AVIF images could strip HDR/color signaling even when pixel data was otherwise preserved.

## Changes

- Read HEIC CICP/NCLX metadata into the `heic:cicp` image property.
- Read HEIC CLLI metadata into the `heic:clli` image property.
- Preserve source CICP on HEIC/AVIF write when available.
- Preserve source CLLI on HEIC/AVIF write when available.
- Add explicit `heic:clli=MaxCLL,MaxFALL` write support.
- Add opt-outs: `heic:preserve-cicp=false` and `heic:preserve-clli=false`.
- Keep explicit `heic:cicp` behavior as the highest priority override.
- Skip implicit CICP preservation when an ICC profile is present, to avoid changing ICC/NCLX interaction.
- Add docs for the new and updated defines and libheif version requirements.
- Add a focused CLI TAP regression test.

## Compatibility Notes

The new libheif calls are version-gated.

CICP/NCLX preservation requires libheif 1.17.0 or later. CLLI preservation requires libheif 1.20.0 or later.

Older libheif builds should continue to compile, with these metadata preservation paths omitted.

This PR does not change HEIC/AVIF chroma selection or downsampling defaults.

## Testing

Tested locally with:

```sh
make -j8
make check TESTS=tests/cli-heic.tap
```

The new TAP test covers source CICP/CLLI readback, resize/transcode preservation, explicit CICP override, explicit CLLI override, `heic:preserve-cicp=false`, `heic:preserve-clli=false`, out-of-range CLLI rejection, and malformed CLLI rejection.

Also tested with a real HDR AVIF resize:

```text
source: 1920 x 1080 | 12/16/6/1 | 1624,182
output:  960 x 540 | 12/16/6/1 | 1624,182
```

## Non-Goals

This PR intentionally does not address unrelated encoder behavior such as default chroma conversion or libheif encoder issues outside ImageMagick's control.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -6195,6 +6195,7 @@ tests_wandtest_LDADD = $(MAGICKCORE_LIBS) $(MAGICKWAND_LIBS)
 TESTS_XFAIL_TESTS = 
 TESTS_TESTS = \
   tests/cli-colorspace.tap \
+  tests/cli-heic.tap \
   tests/cli-pipe.tap \
   tests/cli-svg.tap \
   tests/validate-colorspace.tap \
--- a/coders/heic.c
+++ b/coders/heic.c
@@ -208,6 +208,61 @@ static inline MagickBooleanType IsHEIFSuccess(Image *image,
   return(MagickFalse);
 }
 
+#if LIBHEIF_NUMERIC_VERSION >= HEIC_COMPUTE_NUMERIC_VERSION(1,17,0)
+static void ReadHEICCICPProfile(Image *image,
+  struct heif_image_handle *image_handle,ExceptionInfo *exception)
+{
+  char
+    property[MagickPathExtent];
+
+  struct heif_color_profile_nclx
+    *nclx_profile;
+
+  struct heif_error
+    error;
+
+  nclx_profile=(struct heif_color_profile_nclx *) NULL;
+  error=heif_image_handle_get_nclx_color_profile(image_handle,&nclx_profile);
+  if (error.code == heif_error_Color_profile_does_not_exist)
+    return;
+  if (IsHEIFSuccess(image,&error,exception) == MagickFalse)
+    {
+      if (nclx_profile != (struct heif_color_profile_nclx *) NULL)
+        heif_nclx_color_profile_free(nclx_profile);
+      return;
+    }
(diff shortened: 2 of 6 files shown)


gregbenz:
If anyone needs a test image:
[P3 HDR.avif.zip](https://github.com/user-attachments/files/29355254/P3.HDR.avif.zip)

Transcoding this image demonstrates preservation of wide gamut, HDR, and CLLI metadata.

0fbcb238c0:
magick fails to encode certain AVIFs with this PR.
```
magick: Encoder plugin generated an error: Encoding problem: Subsampling must be 0 with AOM_CICP_MC_IDENTITY. (8.5002)
```

Here's the image, which worked with the current stable release
 I can't seem to attach anything, so here's a link instead.
https://litter.catbox.moe/p598fbbxh2tt9j3k.zip

kmilos:
NB: transcoding to/from PNG should ideally also support preserving these signals in [cICP](https://www.w3.org/TR/png-3/#cICP-chunk) and [cLLI ](https://www.w3.org/TR/png-3/#cLLI-chunk) chunks.

gregbenz:
@kmilos Agreed, as well as for the analogous mechanisms in JXL. Those would be ideal updates if not already well supported (I have not tested those with IM).

PR #8825, opened 2026-06-26
Read Cairo ARGB32 in host byte order so SVG colors are correct on big-endian

### LINKED ISSUE #8596: Colours off when converting SVG on ppc64 w/ librsvg configured
### ImageMagick version

7.1.2-12

### Operating system

Linux

Debian GNU/Linux forky/sid (forky) ppc64

### Description

Converting [`nix-snowflake-colours.svg`](https://github.com/NixOS/nixos-artwork/blob/9d2cdedd73d64a068214482902adea3d02783ba8/logo/nix-snowflake-colours.svg) on my ppc64 machine, with ImageMagick configured with `--with-rsvg` and linked against librsvg [via Nixpkgs](https://search.nixos.org/packages?channel=unstable&query=imagemagick&show=imagemagick), produces wrong colours. See Image 1.

On x86_64-linux, the result of the same command looks correct. See Image 2.

Using librsvg's `rsvg-convert` on the ppc64 machine to do the conversion also produces a correct-looking result, see Image 3. Thus, I'm assuming that this is an issue in ImageMagick, or with its usage of librsvg, but not in librsvg itself.

7.1.2-12 is the first version I've run on this machine, and the issue was already present there. Issue still exists on 7.1.2-15.

### Steps to Reproduce

1. Build ImageMagick for big-endian POWER w/ `--with-rsvg`
2. Use it to convert the linked SVG, i.e. to PNG

### Images

Image 1: `magick nix-snowflake-colours.svg nix-snowflake.png` on powerpc64-linux

Image

Image 2: `magick nix-snowflake-colours.svg nix-snowflake.png` on x86_64-linux

Image

Image 3: `rsvg-convert nix-snowflake-colours.svg -o nix-snowflake.png` (from librsvg) on powerpc64-linux

Image

Scottcjn:
Fixes #8596.

`RenderRSVGImage` reads the Cairo `CAIRO_FORMAT_ARGB32` surface as a fixed `B,G,R,A` byte sequence. ARGB32 is a native-endian 32-bit word (`0xAARRGGBB`), so that byte order is only correct on little-endian hosts. On big-endian the bytes are laid out `A,R,G,B`, so the channels and alpha are misread and SVG colors render incorrectly.

This detects the host byte order once per image and reads `B,G,R,A` on little-endian or `A,R,G,B` on big-endian, using the same `lsb_first` idiom already used in `constitute.c` and `xwd.c`.

### Testing

Built and tested on a native big-endian ppc64 (POWER8) host. Reading an opaque `#ff8040` rect through the rsvg coder (`magick rsvg:test.svg -depth 8 RGBA:-`):

| input | before | after | rsvg-convert reference |
|-------|--------|-------|------------------------|
| `#ff8040` | `ff ff ff ff` | `ff 80 40 ff` | `ff 80 40 ff` |
| red `#ff0000` | white | `255,0,0` | `255,0,0` |
| green `#00ff00` | white | `0,255,0` | `0,255,0` |
| `#c89664` @ 50% | white | `srgba(199,149,100,0.502)` | correct |

The `nix-snowflake-colours.svg` from the issue rendered solid white before this change and renders with the correct colors after.

Little-endian output is unchanged: the little-endian branch reads the same bytes in the same order as before, and the change is confined to the cairo delegate path inside `RenderRSVGImage`.

### DIFF
--- a/coders/svg.c
+++ b/coders/svg.c
@@ -374,6 +374,9 @@ static Image *RenderRSVGImage(const ImageInfo *image_info,Image *image,
   MagickBooleanType
     apply_density;
 
+  EndianType
+    endian;
+
   MemoryInfo
     *pixel_info;
 
@@ -591,6 +594,9 @@ static Image *RenderRSVGImage(const ImageInfo *image_info,Image *image,
       p=gdk_pixbuf_get_pixels(pixel_buffer);
 #endif
       GetPixelInfo(image,&fill_color);
+#if defined(MAGICKCORE_CAIRO_DELEGATE)
+      endian=GetHostEndian();
+#endif
       for (y=0; y < (ssize_t) image->rows; y++)
       {
         q=GetAuthenticPixels(image,0,y,image->columns,1,exception);
@@ -599,15 +605,26 @@ static Image *RenderRSVGImage(const ImageInfo *image_info,Image *image,
         for (x=0; x < (ssize_t) image->columns; x++)
         {
 #if defined(MAGICKCORE_CAIRO_DELEGATE)
-          fill_color.blue=ScaleCharToQuantum(*p++);
-          fill_color.green=ScaleCharToQuantum(*p++);
-          fill_color.red=ScaleCharToQuantum(*p++);
+          if (endian == LSBEndian)
+            {
+              fill_color.blue=ScaleCharToQuantum(*p++);
+              fill_color.green=ScaleCharToQuantum(*p++);
+              fill_color.red=ScaleCharToQuantum(*p++);
+              fill_color.alpha=ScaleCharToQuantum(*p++);
+            }
+          else
+            {
+              fill_color.alpha=ScaleCharToQuantum(*p++);
+              fill_color.red=ScaleCharToQuantum(*p++);
+              fill_color.green=ScaleCharToQuantum(*p++);
+              fill_color.blue=ScaleCharToQuantum(*p++);
+            }
 #else
           fill_color.red=ScaleCharToQuantum(*p++);
           fill_color.green=ScaleCharToQuantum(*p++);
           fill_color.blue=ScaleCharToQuantum(*p++);
-#endif
           fill_color.alpha=ScaleCharToQuantum(*p++);
+#endif
 #if defined(MAGICKCORE_CAIRO_DELEGATE)
           {
             double


coders/svg.c
```diff
@@ -599,15 +609,30 @@ static Image *RenderRSVGImage(const ImageInfo *image_info,Image *image,
...
         {
 #if defined(MAGICKCORE_CAIRO_DELEGATE)
-          fill_color.blue=ScaleCharToQuantum(*p++);
-          fill_color.green=ScaleCharToQuantum(*p++);
-          fill_color.red=ScaleCharToQuantum(*p++);
+          /*
+            Cairo ARGB32 is a native-endian word: the bytes are B,G,R,A on
+            little-endian hosts and A,R,G,B on big-endian hosts (#8596).
+          */
+          if (lsb_first != MagickFalse)
```
dlemstra: You can use `image_info->endian` here instead

coders/svg.c
```diff
@@ -599,15 +609,30 @@ static Image *RenderRSVGImage(const ImageInfo *image_info,Image *image,
         for (x=0; x < (ssize_t) image->columns; x++)
         {
 #if defined(MAGICKCORE_CAIRO_DELEGATE)
-          fill_color.blue=ScaleCharToQuantum(*p++);
-          fill_color.green=ScaleCharToQuantum(*p++);
-          fill_color.red=ScaleCharToQuantum(*p++);
+          /*
+            Cairo ARGB32 is a native-endian word: the bytes are B,G,R,A on
+            little-endian hosts and A,R,G,B on big-endian hosts (#8596).
```
dlemstra: Do not add references to issues in our code. And this comment is pretty redundant so please remove it.

dlemstra (commented):
Thanks for fixing this for us. I left a couple comments for you.

Scottcjn:
Got the comments, resolving now. Have lots of weird hardware so I enjoy resolving these kinds of issues. Thanks for the opportunity.

Scottcjn:
Thanks for the review. Both points addressed in the latest push.

Removed the comment and the issue reference.

On the endian source: `image_info->endian` is the file `-endian` option and defaults to `UndefinedEndian` (it is only set from the endian image option, image.c around line 4170), so it does not carry the host byte order. The Cairo ARGB32 surface is always a host-order word, so the read has to key off the host rather than the file option. Selecting on `image_info->endian` would send little-endian hosts down the big-endian branch by default. I switched the bespoke boolean to an `EndianType` resolved with the same `(*(char *) &x) == 1` host probe already used in constitute.c, mime.c, signature.c and xwd.c, and the branch now reads `if (endian == LSBEndian)`.

Retested on a native big-endian ppc64 host: `#ff8040` reads back `ff 80 40 ff` and the RGB bands are correct. Little-endian output is unchanged.

dlemstra:
Yeah we should not use that. I have added a new method called `GetHostEndian`, can you use that instead?

Scottcjn:
Done, switched to GetHostEndian and rebased onto current main. Much cleaner, thanks for adding it. Retested on a native big-endian ppc64 host: #ff8040 reads back ff 80 40 ff and the RGB bands render correctly. Little-endian output is unchanged.

dlemstra:
Thanks for adding this for us. I made a couple tiny cosmetic changes after the merge.

PR #8826, opened 2026-06-26
reject otb files that declare more pixels than the file holds

dxbjavid:
**Oversized dimensions accepted in the otb reader**
The width and height come straight from the header and are handed to SetImageExtent without any check that the file is large enough to hold that many pixels, so a handful of bytes can make the reader size the pixel cache and walk billions of one-bit pixels before the decode runs out of input. This mirrors the insufficient-data guard wbmp recently received, placed just before the extent is set.


dxbjavid:
Thank you for the review and merge. Appreciated

PR #8828, opened 2026-06-28
Fix HEIC identity CICP with chroma subsampling

gregbenz:
## Summary

Follow-up to #8824.

This fixes the AVIF encoding failure reported in https://github.com/ImageMagick/ImageMagick/pull/8824#issuecomment-4826903309:

```text
Subsampling must be 0 with AOM_CICP_MC_IDENTITY
```

The issue occurs when a source image has preserved CICP matrix coefficient `0` (RGB/GBR identity matrix) and ImageMagick/libheif attempts to encode subsampled AVIF output.

## Fix

This keeps CICP preservation enabled while avoiding invalid matrix/subsampling combinations:

- If preserved source CICP uses matrix coefficient `0` and `heic:chroma` is omitted, ImageMagick uses `heic:chroma=444` so matrix `0` remains valid.
- If preserved source CICP uses matrix coefficient `0` and the user explicitly requests `heic:chroma=420` or `heic:chroma=422`, ImageMagick sets libheif's output nclx profile to matrix coefficient `6` (BT.601) for the subsampled YCbCr output.
- If the user explicitly requests `heic:cicp=.../0/...` together with `heic:chroma=420` or `heic:chroma=422`, ImageMagick now fails with a clear option error instead of writing inconsistent metadata.

This keeps default behavior user-friendly while still rejecting an explicitly invalid request.

## Validation

I reproduced the reported failure against the code merged in #8824 using an AVIF source with CICP `1/13/0/1`.

With the old code, both default resize and explicit `heic:chroma=420` failed with:

```text
Subsampling must be 0 with AOM_CICP_MC_IDENTITY
```

With this patch:

- default resize succeeds and preserves `1/13/0/1` as RGB
- `heic:chroma=420` succeeds and writes `1/13/6/1` as YCbCr 4:2:0
- `heic:chroma=422` succeeds and writes `1/13/6/1` as YCbCr 4:2:2

## Tests

Added/updated `tests/cli-heic.tap` coverage for:

- identity matrix CICP preservation
- default identity matrix handling
- explicit `heic:chroma=444`
- requested `heic:chroma=420`
- requested `heic:chroma=422`
- explicit invalid matrix `0` CICP with subsampling
- `heic:preserve-cicp=false` with subsampling

Local verification:

```text
git diff --check
make -j8
make check TESTS=tests/cli-heic.tap
```

### DIFF
--- a/coders/heic.c
+++ b/coders/heic.c
@@ -1407,8 +1407,76 @@ static void WriteProfile(struct heif_context *context,Image *image,
 }
 
 #if LIBHEIF_NUMERIC_VERSION >= HEIC_COMPUTE_NUMERIC_VERSION(1,17,0)
-static MagickBooleanType WriteHEICCICPProfile(Image *image,
-  struct heif_image *heif_image,const char *option,ExceptionInfo *exception)
+static const char *GetHEICCICPOption(const ImageInfo *image_info,Image *image,
+  ExceptionInfo *exception)
+{
+  const char
+    *option;
+
+  option=GetImageOption(image_info,"heic:cicp");
+  if ((option == (const char *) NULL) &&
+      (GetImageProfile(image,"icc") == (const StringInfo *) NULL) &&
+      (IsStringFalse(GetImageOption(image_info,"heic:preserve-cicp")) ==
+       MagickFalse))
+    option=GetImageProperty(image,"heic:cicp",exception);
+  return(option);
+}
+
+static MagickBooleanType IsHEICCICPIdentityMatrix(const char *option)
+{
+  GeometryInfo
+    cicp;
+
+  MagickStatusType
+    flags;
+
+  if (option == (const char *) NULL)
+    return(MagickFalse);
+  SetGeometryInfo(&cicp);
+  flags=ParseGeometry(option,&cicp);
+  if (((flags & XiValue) != 0) && (cicp.xi == 0.0))
+    return(MagickTrue);
+  return(MagickFalse);
+}
+
+static MagickBooleanType IsHEICExplicitCICP(const ImageInfo *image_info)
+{
+  if (GetImageOption(image_info,"heic:cicp") != (const char *) NULL)
+    return(MagickTrue);
+  return(MagickFalse);
+}
+
+static MagickBooleanType IsHEICSubsampledChroma(const char *option)
+{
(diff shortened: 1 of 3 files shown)


coders/heic.c
```diff
@@ -1859,6 +1985,11 @@ static MagickBooleanType WriteHEICSequenceImage(const ImageInfo *image_info,
   struct heif_image
     *heif_image = (struct heif_image *) NULL;
 
+#if LIBHEIF_NUMERIC_VERSION >= HEIC_COMPUTE_NUMERIC_VERSION(1,17,0)
```
dlemstra: You don't need all these version `1.17.0` checks in this method. The whole `WriteHEICSequenceImage` method is wrapped inside a version `1.20.0` check.

coders/heic.c
```diff
@@ -1923,10 +2060,25 @@ static MagickBooleanType WriteHEICSequenceImage(const ImageInfo *image_info,
...
     }
   seq_options=heif_sequence_encoding_options_alloc();
-#if LIBHEIF_NUMERIC_VERSION >= HEIC_COMPUTE_NUMERIC_VERSION(1,21,0)
   if (seq_options != (struct heif_sequence_encoding_options *) NULL)
-    seq_options->save_alpha_channel=1;
+    {
+#if LIBHEIF_NUMERIC_VERSION >= HEIC_COMPUTE_NUMERIC_VERSION(1,21,0)
+      seq_options->save_alpha_channel=1;
+#endif
+#if LIBHEIF_NUMERIC_VERSION >= HEIC_COMPUTE_NUMERIC_VERSION(1,17,0)
```
dlemstra: Please swap the order here, version `1.17.0` stuff before `1.21.0`.

gregbenz: Thanks, addressed both in the latest update.

I removed the redundant `1.17.0` guards inside `WriteHEICSequenceImage()` and moved the CICP/output profile setup before the `1.21.0` `save_alpha_channel` assignment.

Local verification:
- `git diff --check`
- `make -j8 utilities/magick`
- `make -j8`
- `make check TESTS=tests/cli-heic.tap` passed 14/14
- manual two-frame AVIF sequence write/read smoke test passed

dlemstra (approved):

PR #8832, opened 2026-06-30
guard ept length overflow before allocation

dxbjavid:
ReadEPTImage takes the postscript and tiff section lengths straight from the header and passes length+1 to AcquireQuantumMemory. On a build where size_t is 32-bit a length close to UINT_MAX wraps the +1 to a tiny allocation that the following read then overruns. This adds the same UINT_MAX guard that was just added to the jnx reader so an oversized length is rejected before the allocation.


urban-warrior:
We added a check for overflow with the HeapOverflowCheckAdd() macro a few weeks ago.  If you agree with our approach, feel free to close this pull request..

dxbjavid:
Agreed, HeapOverflowCheckAdd(length,1) on both sections covers the same case and it fires before the allocation, which is the part that mattered. It also avoids hardcoding UINT_MAX the way I did, which wouldn't have been the right bound on a 64-bit size_t anyway. Looks like it landed in 92b961a before I opened this, and I'd based the branch on an older tree, so I missed it. Closing.

PR #8920, opened 2026-08-22
Fix inline SVG style precedence

### LINKED ISSUE #8799: Some mono SVGs gain colours when converted.
### ImageMagick version

7.1.2.18

### Operating system

Linux

Ubuntu 26.04

### Description

Look at this Mono, black and white image of Tux the pengin
https://upload.wikimedia.org/wikipedia/commons/3/3a/Tux_Mono.svg
When displayed in Firefox, or opened in GIMP, it's clearly just black and white.

But when I try 
`convert Tux_Mono.svg tux.png`
 the result comes up with a yellow beak and a yellow left foot.

(the same thing happens when I use `display Tux_Mono.svg`.

eog, qiv, inkscape and chromium also show the SVG in black/white.

So convert/display are the odd ones out - but if I feed the file through "svgo", then the optimised svg behaves correctly in all cases.

### Steps to Reproduce

Download the file from wikipedia.
Open it with "display" 
Note that it renders differently in ImageMagick than every other tool.

### Images

https://upload.wikimedia.org/wikipedia/commons/3/3a/Tux_Mono.svg

arhxam:
The SVG specifications give inline style declarations higher priority than
presentation attributes. The MSVG parser currently applies both in XML
attribute order, so a presentation attribute can incorrectly override an
earlier `style` attribute.

This defers the existing inline-style processing until after presentation
attributes have been handled. The regression test covers both attribute orders
and confirms that inline CSS wins in either case.

Fixes #8799.

Tests: `make check` (52 passed, 1 delegate-dependent test skipped)

AI assistance was used to investigate the parser and draft the change. I
reviewed the diff and ran the tests and reproductions described above locally.

### DIFF
--- a/tests/cli-svg.tap
+++ b/tests/cli-svg.tap
@@ -24,6 +24,11 @@ echo "1..2"
 ${MAGICK} ${SRCDIR}/input_svg_gradient_transform.svg null: && echo "ok" || echo "not ok"
 
 # Inline styles override presentation attributes regardless of attribute order.
+msvg_mode=`${MAGICK} -list format | awk '$1 ~ /^MSVG/ { print $3 }'`
+case "$msvg_mode" in
+  *r*) ;;
+  *) echo "ok # SKIP MSVG coder unavailable"; exit 0 ;;
+esac
 colors=`${MAGICK} MSVG:${SRCDIR}/input_svg_inline_style.svg \
   -format '%[fx:int(255*p{5,5}.r+0.5)],%[fx:int(255*p{5,5}.g+0.5)],%[fx:int(255*p{5,5}.b+0.5)] %[fx:int(255*p{15,5}.r+0.5)],%[fx:int(255*p{15,5}.g+0.5)],%[fx:int(255*p{15,5}.b+0.5)]' info:-`
 [ "X$colors" = "X255,255,255 255,255,255" ] && echo "ok" || echo "not ok"
--- a/tests/cli-svg.tap
+++ b/tests/cli-svg.tap
@@ -24,12 +24,13 @@ echo "1..2"
 ${MAGICK} ${SRCDIR}/input_svg_gradient_transform.svg null: && echo "ok" || echo "not ok"
 
 # Inline styles override presentation attributes regardless of attribute order.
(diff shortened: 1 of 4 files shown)


dlemstra (dismissed):

dlemstra:
Your test seems to fail in one of our builds. Can you check why?

arhxam:
The GCC job is configured without the XML delegate, so its MSVG registration is write-only (`-w+`) and the new internal-renderer test cannot decode the fixture there. I pushed a capability-based skip for that exact case; builds where MSVG has read support still run the pixel assertion. I also verified the TAP script syntax and both `rw+`/`-w+` detection paths locally. The updated CI run should confirm the matrix.

arhxam:
Additional verification: I configured this branch with `--without-xml --without-rsvg`, built `utilities/magick`, and ran `make check TESTS=tests/cli-svg.tap`. Result: 1 pass, 1 explicit `MSVG coder unavailable` skip, 0 failures. The replacement GitHub run is currently marked `action_required` pending maintainer approval.

tests/cli-svg.tap
```diff
@@ -18,8 +18,18 @@
...
+echo "1..2"
 
 # gradientTransform should not cause a double-free / SIGABRT (#8582)
 ${MAGICK} ${SRCDIR}/input_svg_gradient_transform.svg null: && echo "ok" || echo "not ok"
+
+# Inline styles override presentation attributes regardless of attribute order.
+msvg_mode=`${MAGICK} -list format | awk '$1 ~ /^MSVG/ { print $3 }'`
+case "$msvg_mode" in
+  *r*) ;;
+  *) echo "ok # SKIP MSVG coder unavailable"; exit 0 ;;
```
dlemstra: An exit here will cause issues when another test is added.

arhxam: Fixed in e13ca4ac5. The capability check now reads the mode column, runs the pixel assertion only when MSVG is readable, and emits the skip without exiting, so later TAP tests continue. Verification: the XML-enabled build passed 2/2 SVG tests; the `--without-xml --without-rsvg` build reported 1 pass and 1 explicit skip; an exact-head control-flow check also confirmed a following test still runs after the skip.

dlemstra (approved):

PR #8921, opened 2026-08-22
Preserve page offsets in PCX images

### LINKED ISSUE #8847: Round-trip PCX and TGA (X,Y) offsets.
### Is your feature request related to a problem? Please describe.

Both the .pcx and the .tga formats read but ignore the XOffset, YOffset (or equivalently Left, Top) fields.

ImageMagick supports these values via the Page, and so the values should round-trip.

### Describe the solution you'd like

Each format should round-trip their own coordinates.  The coordinates can be converted to X, Y, Width, Height for representation as a Page.

### Describe alternatives you've considered

There really isn't an alternative, ImageMagick is throwing away necessary data.

### Additional context

Many, many PlayStation image files contain X,Y offsets.

arhxam:
PCX headers store the image origin in their `left` and `top` fields, but the
decoder currently discards those values and the encoder always writes zero.
This preserves representable page offsets in both directions and calculates
the absolute `right`/`bottom` endpoints without overflowing the 16-bit header.

The regression coverage checks a normal round trip, the maximum representable
endpoint, and the safe fallback for an origin whose endpoint would overflow.
The TGA portion of #8847 has already landed on `main`, so this completes the
remaining PCX portion.

Fixes #8847.

Tests:
- `make check TESTS=tests/cli-pcx.tap` (3 passed)
- `make check` (51 passed, 1 delegate-dependent test skipped)

AI assistance was used to investigate the coder and draft the change. I
reviewed the offset arithmetic, inspected the encoded headers, and ran the
tests and round-trip reproductions locally.

### DIFF
--- a/coders/pcx.c
+++ b/coders/pcx.c
@@ -341,6 +341,8 @@ static Image *ReadPCXImage(const ImageInfo *image_info,ExceptionInfo *exception)
       ThrowPCXException(CorruptImageError,"ImproperImageHeader");
     image->columns=(size_t) (pcx_info.right-pcx_info.left)+1UL;
     image->rows=(size_t) (pcx_info.bottom-pcx_info.top)+1UL;
+    image->page.x=(ssize_t) pcx_info.left;
+    image->page.y=(ssize_t) pcx_info.top;
     image->depth=pcx_info.bits_per_pixel;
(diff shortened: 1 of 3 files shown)


dlemstra (approved):

PR #8935, opened 2026-09-02
fix: use registered UHDR module name in policy check

gregbenz:
### Description

The UHDR module registration was recently updated from `UHDr` to `UHDR`. This updates the corresponding JPEG auto-routing policy check to use the current module name.

Policy matching is case-sensitive, so the previous name could prevent UHDR routing under a deny-by-default module policy that explicitly permits `UHDR`.

### Validation

Using a module build with libultrahdr 1.4.0 and an ISO gain-map JPEG:

- Before this change, a policy allowing `JPEG` and `UHDR` produced `format=JPEG`.
- After this change, the same policy produced `format=UHDR`.
- Omitting `UHDR` from the allowlist continued to produce `format=JPEG`.

This change does not modify default policies, dependency discovery, JPEG detection, or AVIF behavior.


gregbenz:
Thank you for a quick review!

PR #8937, opened 2026-09-03
Preserve EXIF dimensions during transforms

gregbenz:
# Preserve EXIF dimensions during transforms

## Summary

Update EXIF geometry metadata when image dimensions change so transformed images do not retain stale dimensions. This problem was identified in [#8743](https://github.com/ImageMagick/ImageMagick/issues/8743).

## Changes

- Update IFD0 `ImageWidth` and `ImageLength`.
- Update ExifIFD `PixelXDimension` and `PixelYDimension`.
- Preserve IFD1 thumbnail dimensions, orientation, and resolution metadata.
- Preserve TIFF byte order and the existing dimension field type when possible.
- Promote `SHORT` dimensions to `LONG` when required.
- Ignore dimensions that cannot be represented as EXIF `LONG` values.
- Validate complete IFDs and sub-IFD offsets before traversing them.
- Read EXIF tag IDs as unsigned 16-bit values so the ExifIFD pointer and pixel dimension tags above `0x7fff` are recognized.

## Tests

The EXIF regression test covers:

- little-endian and big-endian profiles;
- `SHORT` and `LONG` dimensions in IFD0 and ExifIFD;
- `SHORT` to `LONG` promotion at 65,536 pixels;
- preservation of IFD1 thumbnail dimensions, orientation, and resolution;
- malformed ExifIFD pointers and incomplete IFDs, including an invalid IFD1 after valid primary metadata; and
- builds where the optional PNG coder is unavailable.

All 11 focused assertions pass in builds configured against libultrahdr 1.4.0 and 2.0.2. A resized copy of the sample from #8743 reports matching 540x540 image, IFD0, and ExifIFD dimensions.

### DIFF
--- a/MagickCore/profile.c
+++ b/MagickCore/profile.c
@@ -2302,14 +2302,36 @@ static void WriteProfileShort(const EndianType endian,
   (void) memcpy(p,buffer,2);
 }
 
+static inline void WriteProfileDimension(const EndianType endian,
+  const size_t dimension,const ssize_t format,const int components,
+  unsigned char *entry,unsigned char *value)
+{
+  if ((components != 1) || ((format != 3) && (format != 4)))
+    return;
+  if ((format == 3) && (dimension <= MAGICK_USHORT_MAX))
+    {
+      WriteProfileShort(endian,(unsigned short) dimension,value);
+      return;
+    }
+  if (dimension > UINT32_MAX)
+    return;
+  if (format == 3)
+    WriteProfileShort(endian,4,entry+2);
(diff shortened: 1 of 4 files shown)


dlemstra:
Lets not mix this in a single PR. Please split this.

gregbenz:
@dlemstra Thank you for the quick guidance. I have updated this PR to focus on the EXIF dimensions only.

I created #8938 to separately help ensure small gain map values don't get set to 0.

PR #8947, opened 2026-09-08
Honor gain-map quality when transforming Ultra HDR images

gregbenz:
Resizing an Ultra HDR JPEG can ignore the requested gain-map quality. When ImageMagick transforms and re-encodes the gain map, it uses the map's original JPEG quality instead of `uhdr:gainmap-quality`, or `-quality` when no override is provided. This prevents callers from controlling the map's compression and can leave the output larger or more compressed than intended.

Set the chosen quality on the gain-map image that the JPEG encoder reads. When no gain-map transform is needed, the existing compressed map is still reused.

The regression compares two resized outputs with the same base-image quality and different gain-map quality settings. It checks that the SDR pixels stay identical while the decoded HDR pixels differ. All 18 focused assertions pass with libultrahdr 1.4.0 and 2.0.2 on macOS ARM64. Separate checks of the extracted gain-map JPEGs confirm the explicit quality and the ordinary-quality fallback.

### Prerequisites


urban-warrior:
@gregbenz, this pull request has a conflict.  Resolve the conflict before we can review and merge.  Thanks.

gregbenz:
@urban-warrior Thanks, should be ready now.  All 18 focused UHDR assertions pass locally with libultrahdr 1.4.0 and 2.0.2. CI in progress.

